PR #438, opened 2013-06-14
Javascript and unittest fixes

### LINKED ISSUE #436: Patch 'contains' and 'startsWith'
OpenLayers upgrade broke the code in some browsers by not including support for these methods.

notthatbreezy:
This pull request fixes #436 by creating a new Javascript utility file that implements the string methods `contains` and `startsWith` that were broken after an OpenLayers upgrade. Additionally, it replaces uses of these methods in Javascript files.

Finally, unittests were failing due to 2 minor issues and were fixed:
1. Float comparisons now use `assertAlmostEqual` instead of `assertEqual`
2. Template tests were failing due to some changes in html by designers, these were modified

### DIFF
--- a/django/publicmapping/redistricting/static/js/chooseplan.js
+++ b/django/publicmapping/redistricting/static/js/chooseplan.js
@@ -127,7 +127,7 @@ chooseplan = function(options) {
      */
     var selectPlan = function () {
         // Notify the reaggregator of reaggregation button clicks
-        if (db_util.startsWith($('#start_mapping .ui-button-text').html(), 'Reaggregate')) {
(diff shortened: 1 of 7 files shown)


kshepard:
Looks good. Minor change request: I foresee adding other javascript utility files like this. I think we should add a root 'DB' namespace object and keep the functionality within there in order to minimize our global namespace impact. 

So in place of the initialization of db_util do something like:

var DB = {};
DB.util = {};
...etc...

And then we'll make calls to DB.util.startsWith, etc.

notthatbreezy:
These 2 commits deal with the suggestion for changing the db_util to DB.util

PR #439, opened 2013-06-24
Fix Admin Password Change Bug

### LINKED ISSUE #423: Admin Password Reset Error
The admin password reset does not appear to be functioning for the Mexico test instance. I verified with an account called "password_test_reset" Once the password has been changed through the admin account, neither the old and newly assigned passwords do not work.

notthatbreezy:
This fixes #423.

The admin page was trying to load jquery from the wrong path. This caused the bug because there is javascript on the page that hashes the password before sending off to the server. Without jquery the password wasn't being hashed and being sent as the plain text to the server.

Changing the path that loads javascript to the correct location fixes the problem.

### DIFF
--- a/django/publicmapping/management/commands/changepassword.py
+++ b/django/publicmapping/management/commands/changepassword.py
@@ -1,0 +1,64 @@
+import getpass
+from optparse import make_option
+
+from hashlib import sha1
+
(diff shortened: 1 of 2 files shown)


kshepard:
lgtm

PR #440, opened 2013-07-19
Change download plan type from shape to index file

kshepard:


wboykinm:
I realize I missed this until just now, but seriously? What good is an index file to anyone's workflow? This adds extra steps to get an export of valid geodata. If the weird .shp file structure was a problem, why not change it to .kml or .geojson export?

kshepard:
@wboykinm - In the last round of QA, @maltman noticed that this had been changed to a shapefile (it was originally an index file, and seems to have been swapped along the way, most likely by accident). The reasoning for reverting it to an index file is to maintain consistency: the plan chooser page only accepts plans uploaded as index files, so it seemed strange to be able to upload a file in one format, but then download it in a different format. Note that this change only applies to the plan chooser page -- plans can still be exported as shapefiles via the 'Export' dropdown on the plan page.

wboykinm:
This is qualified by the fact that I'm only basing this on [a single deployment](http://btvvotes.org), but our experience was of many more users downloading plans for use in their own analyses than of users uploading plans. In fact the only plan uploaded at any point was a single revision to the existing census numbers. I would almost call upload and download two different components of the UX. I understand the intention of consistency, but in this case the "shapefile download" option was actually sort of a killer feature for the GIS crowd.

rcheetham:
Bill,

As someone who comes to redistricting from the GIS direction, I don't
disagree with the shapefile being useful and a killer feature.  But in the
redistricting world, there is a similar group of people who use these index
files to move district plans between different systems.  I think the ideal
would probably be to provide both in each context in which it might be
useful.

Robert

On Friday, July 19, 2013, Bill Morris wrote:

> This is qualified by the fact that I'm only basing this on a single
> deployment http://btvvotes.org, but our experience was of many more
> users downloading plans for use in their own analyses than of users
> uploading plans. In fact the only plan uploaded at any point was a single
> revision to the existing census numbers. I would almost call upload and
> download two different components of the UX. I understand the intention of
> consistency, but in this case the "shapefile download" option was actually
> sort of a killer feature for the GIS crowd.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/PublicMapping/DistrictBuilder/pull/440#issuecomment-21279930
> .

## 

---

Robert Cheetham

Azavea  |  340 N 12th St, Ste 402, Philadelphia, PA
cheetham@azavea.com  | T 215.701.7713  | F 215.925.2663
Web azavea.com http://www.azavea.com/  |  Blog azavea.com/blogs  |
Twitter @ http://goog_858212415rcheetham http://twitter.com/rcheetham and
@azavea http://twitter.com/azavea

_Azavea is a B Corporation http://www.bcorporation.net/what-are-b-corps -
we apply geospatial technology to create better communities *
*while advancing the state-of-the-art through research. Join us in creating
a better world._

wboykinm:
That's good to know the index files see traffic elsewhere. Indeed, I'm not really equipped to answer the question of whether a shapefile or an index file is more conducive to accessibility by the lay public. I'm inclined to say that neither of them are. Perhaps it would be best to have index files available [in the "Plan" tab](http://flic.kr/p/fddzyk) while keeping [multiple options in "View"](http://www.flickr.com/photos/vtcraghead/9330027458/)? I haven't deployed the trunk version here to know if that's what this merge accomplished.
I may simply be missing a shift in the districtbuilder use case: is it not serving as a tool for the general public anymore? Is it now marketed more as a Maptitude replacement with an edge toward process transparency? I would call the two functions equally valuable, though the former is certainly harder to pull off.
Apologies for getting existential because of a simple toolbar switch, and thanks for developing the platform in the first place. It's been invaluable.

rcheetham:
Bill,

Those are good ideas.  I'll talk to the development team about how feasible
this is given our limited resources.

I don't think we've changed the marketing for DistrictBuilder.  It's always
been aimed at both public use as well as use by legislatures for both
constructing and sharing plans with the public.

Best,

Robert

---

Robert Cheetham

Azavea  |  340 N 12th St, Ste 402, Philadelphia, PA
cheetham@azavea.com  | T 215.701.7713  | F 215.925.2663
Web azavea.com http://www.azavea.com/  |  Blog azavea.com/blogs  |
Twitter @ http://goog_858212415rcheetham http://twitter.com/rcheetham and
@azavea http://twitter.com/azavea

_Azavea is a B Corporation http://www.bcorporation.net/what-are-b-corps -
we apply geospatial technology to create better communities *
*while advancing the state-of-the-art through research. Join us in creating
a better world._

kshepard:
Bill,

Earlier today we merged pull request #446 into master. This alters the plan chooser page to display two download buttons: one for a shapefile and one for a district index file. Thank you very much for providing feedback and helping to improve DistrictBuilder's user experience.

Best regards,
Kenny

wboykinm:
Thanks for listening!

PR #443, opened 2013-08-27
Adjacency and Convex Hull Calculator Edits

notthatbreezy:
- Fix redis/python memory issue when calculating adjacency
- Add choropleth option for adjacency and convex hull calculators
- Make adjacency and convex hull choropleths optional


kshepard:
+1

Issue #460, opened 2017-10-03
Dig through template inheritance to fix lookups


jisantuc:
Something is causing django to get confused about what views or templates it's supposed to be looking for.

```
Reverse for '' not found. '' is not a valid view function or pattern name.
```

The first place this happens is when you click the login button.

This task is to fix this issue on the login button and breadth-first search from there to try to find more instances where this needs to be fixed, ideally identifying something easily searchable so it can be fixed everywhere.

Issue #459, opened 2017-10-03
Fix translations


jisantuc:
On the landing page, javascript complains that it doesn't know what `gettext` is:

```
Uncaught ReferenceError: gettext is not defined
```

It needs to know what that is for translations. Make it understand.

PR #455, opened 2017-08-29
WIP: Deploy with Docker

themightychris:

PR #461, opened 2017-10-03
Feature/js/setup dev environment

jisantuc:
Overview
-----

This PR takes steps toward having a working local instance of the application.

Its main actions are:

- run the django server using gunicorn
- remove unused dependencies
- move files around as needed
- move the django server into a container
- move the database and redis into containers
- upgrade a bunch of dependencies (django, redis, django-redis, postgres, postgis...)
- use django tools that may not have existed in django 1.4 (like `django.core.cache` and the `{% static ... %}` interpolator)
- upgrade django settings (include a `CACHES` section for the cache, e.g.)
- modify url routing
- kill a _ton_ of trailing whitespace (but by no means all of it -- it's dead everywhere I visited I think but I probably forgot to do it in some places)

It's _huge_, and therefore pretty much impossible to test completely.

In the application's current state, you can hit the landing page and all of the links appear to work, but trying to do (basically) anything fails. Pretty much the only thing that works is that you can try to log in as a guest, and then the app will successfully tell you that you've failed to login (:fire:). This prevents exercising some of the application's features.

Notes
-----

The "spread firefox support" button seems not to exist anywhere in mozilla anymore. I didn't remove it because I wasn't sure if we wanted to go hunting for a replacement.

There are a few things that I either haven't exercised or haven't fixed:

- geoserver remains untested. I'll integrate the container into this work in when I get a PR up for it
- `gettext` isn't working at all. That seems really bad! (#459)
- Something is generically wrong with basically all of the views, where they complain about `No reverse lookup found` or believe they have syntax errors in the templates or who knows what else. Digging through template inheritance seemed like a possible black hole (#460)

Testing
-----

- `docker-compose up`
- Test the cache connection:
  - `docker-compose exec django bash`
  - `./manage.py shell`
  - run this:
```python
from django.core.cache import caches
cache = caches['default']
cache.get('foo')
```
    
- you should get nothing instead of an error
- Test the database connection
  - from inside the django container, `./manage.py dbshell`
  - `\d`
  - you should see a bunch of tables
  - `\d redistricting_district`
  - you should see something that looks like a reasonable data model
- test the frontend
  - navigate to `localhost:8080`
  - the only error you should see in the console should be that the frontend doesn't know what `gettext` is

Closes (or at least convincingly starts on) #455 
Closes #458


django/publicmapping/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM python:2.7
```
hectcastro: This overlaps a good bit with what's in `quay.io/azavea/django`. I think that it would be good to use that as the base, or at least split that out as a separate task for the future since this PR is already pretty big.

See: https://github.com/azavea/docker-django/blob/master/Dockerfile-slim.template

django/publicmapping/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        gdal-bin \
+        gettext \
+        postgresql-client=${PG_VERSION} \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY requirements.txt /usr/src/app/
+RUN pip install --no-cache-dir \
+        numpy==$(grep "numpy" requirements.txt | cut -d= -f3) \
+        scipy==$(grep "scipy" requirements.txt | cut -d= -f3) \
```
ddohler: On `docker-compose up` this is throwing an error:
```
Collecting numpy==1.6.2
  Downloading numpy-1.6.2.tar.gz (2.6MB)
Collecting scipy==0.11.0
  Downloading scipy-0.11.0.tar.gz (6.8MB)
    Complete output from command python setup.py egg_info:
      File "<string>", line 20, in <module>
      File "/tmp/pip-build-aIUO1B/scipy/setup.py", line 208, in <module>
        setup_package()
      File "/tmp/pip-build-aIUO1B/scipy/setup.py", line 145, in setup_package
        from numpy.distutils.core import setup
    ImportError: No module named numpy.distutils.core
```

django/publicmapping/publicmapping/settings.py
```diff
@@ -0,0 +1,199 @@
...
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/1.11/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = '1%a60e98ywm#$gl4mcm%s#uxoppc3)t%w+an*gzmujo-v-i07^'
```
ddohler: It might be good to change this or deliberately leave it out so that it doesn't slip through the cracks as we work on this further.

django/publicmapping/redistricting/urls.py
```diff
@@ -22,53 +22,55 @@
...
-    (r'^processingstatus/$', 'get_processing_status'),
-    (r'^feed/plans/$', 'plan_feed'),
-    (r'^feed/shared/$', 'share_feed'),
-    (r'^i18n/', include('django.conf.urls.i18n')),
-)
+from .  import views as redistricting_views
+
+urlpatterns = [
+    url(r'^plan/$', redistricting_views.viewplan, { 'planid': 0 }),
+    url(r'^plan/create/$', redistricting_views.createplan),
```
ddohler, line 36: The removal of the `name` parameters here may be the source of the confusion with `reverse()` if we were relying on calling these views by name. In other words, this line may need to become `url(r'^plan/create/$', redistricting_views.createplan, name='createplan')`.

ddohler (changes requested):
`docker-compose up` threw errors for me that prevented me from moving further. Otherwise this looks fine (although I mainly stuck to reading the parts that looked hand-written and skimmed the rest of it).

django/publicmapping/publicmapping/settings.py
hectcastro: Yeah, I think that we can pull it from the environment such that it leaves this setting unset if we forget to populate it:

```python
SECRET_KEY = os.getenv('DJANGO_SECRET_KEY')
```

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: "3.3"
+services:
+  postgres:
+    image: quay.io/azavea/postgis:2.2-postgres9.5-slim
+    environment:
+      - POSTGRES_DB=district_builder
+      - POSTGRES_USER=district_builder
+      - POSTGRES_PASSWORD=district_builder
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready"]
```
hectcastro: We've seen instances of this failing when the service is up because of the missing `-U` argument to `pg_isready`. Error looks like the following in the logs:

```
FATAL:  role "root" does not exist
FATAL:  role "root" does not exist
FATAL:  role "root" does not exist
...
```

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+      interval: 15s
+      timeout: 5s
+      retries: 3
+
+  redis:
+    image: redis:3.2.11-alpine
```
hectcastro: I think that we can shorten this to `3.2-alpine`, although, would be great to get rid of Redis in the future if we can.

django/publicmapping/Dockerfile
hectcastro: Hm. I did not seem to run into this after running:

```bash
$ docker-compose build
$ docker-compose up
```

hectcastro:
This was a large set of changes, so possibly something to do as a follow-up task based on how many points were allocated, but I think that it would be good to get the STRTA setup going before other people start picking this up.

This commit is a good place to start from. It has way more that necessary, so I'd focus on only the Vagrant, Ansible, `scripts` parts that we're going to use in the immediate future.

django/publicmapping/Dockerfile
ddohler: I nuked every docker image on my machine and then rebuilt, and now it works. ¯\_(ツ)_/¯

docker-compose.yml
ddohler: I can confirm that I'm seeing this after doing `docker-compose up`.

ddohler:
A couple more things that I noticed once I got the docker containers building:
- It would be good to install `ipython` and `django-extensions` into the container so that we have access to `./manage.py shell_plus`.
- When I ran `./manage.py dbshell` I got a warning about Postgresql client / server version mismatches: `WARNING: psql major version 9.4, server major version 9.5. Some psql features might not work.`

docker-compose.yml
jisantuc: I saw the error, but it didn't appear to affect anything -- I could still fire up a django shell and talk to the database so I thought of it as just an annoying thing. Fixing.

jisantuc:
I think we talked about not doing the Vagrant + docker setup that we normally do to decrease friction with community contributors. Am I remembering that wrong?

docker-compose.yml
jisantuc: There's no need to use redis here I think, I just did because the project used redis previously. Do you mean it would be good to switch to memcached, or it would be good not to have to depend on a cache at all?

docker-compose.yml
jisantuc: This doesn't appear to be resolved by including `"-U", "district_builder" in the healthcheck command, though it does show up every 15 seconds, matching the healthcheck interval

django/publicmapping/redistricting/urls.py
jisantuc: I tried adding names to all of the url views including the login view with no change, though I may not have understood correctly what the names should have been

jisantuc:
The warning about client/server version mismatches from `dbshell` is resolved after switching the base image

docker-compose.yml
kshepard: We're actually not using Redis as a cache, but rather as key/value storage for adjacency information, which is used in a single, and rarely used, calculator. If we can find an alternate efficient way to perform that calculation, then Redis can be eliminated as a dependency. At this stage, we can probably just ignore dealing with Redis completely if we just don't use that calculator in any testing.

kshepard:
re: Docker only vs. Vagrant + Docker -- there was some discussion, but I don't think there was ever a definitive decision. We'll need to talk it through.

docker-compose.yml
hectcastro: My comment was mostly from the perspective of eliminating an architectural component. I'm cool with keeping it out for now based on Kenny's explanation.

hectcastro:
My vote would be to include the Vagrant based Docker setup with STRTA scripts for internal consistency. Others can choose to make use of that framework, or not. Even if they don't, I think it would be valuable to surface the steps we take to get a development environment going in code. I also think that including the scripts helps minimize the documentation burden (run `setup` vs. run migrations, load this data, build the containers, etc.).

jisantuc:
Ok scripts are in a place where you can `./scripts/` your way through setup now, and there's a `scripts/test` that lints the scripts directory and does nothing else except hang out and wait for more interesting testing and a CI environment to run in

deployment/ansible/district-builder.yml
```diff
@@ -0,0 +1,14 @@
...
+  pre_tasks:
+    - name: Update APT cache
+      apt: update_cache=yes cache_valid_time=3600
+
+  roles:
+    - { role: "azavea.ntp" }
+    - { role: "district-builder.environment" }
+    - { role: "azavea.python-security" }
+    - { role: "district-builder.docker" }
+    - { role: "district-builder.aws-cli" }
```
jisantuc: My assumption is that we're going to include AWS integration in the repo, but if that's not the case and we're just going to have containers and handle deployment of those containers separately, this can be removed. Currently the aws cli isn't being used for anything so it's a no-cost deletion.

themightychris:
Is STRTA some internal Azavea convention?

Is there any major work left before this considered just about merge ready? I'm eyeing a good point to take another stab at contributing a habitat plan. I think y'all will loooove what it does, it can render all the other debates moot.

jisantuc:
It's not us, it's short for "scripts to rule them all": https://github.com/github/scripts-to-rule-them-all

deployment/ansible/district-builder.yml
hectcastro: We can keep it out for now.

deployment/ansible/group_vars/all
```diff
@@ -0,0 +1,7 @@
+---
+
+aws_cli_version: "1.11.*"
```
hectcastro: Same with the removal of these two.

scripts/update
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+set -e
+
+if [[ -n "${RF_DEBUG}" ]]; then
```
hectcastro: Stale variable name here.

scripts/test
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+
+if [[ -n "${RF_DEBUG}" ]]; then
```
hectcastro: Stale variable name here too.

hectcastro (approved):
A good start. Merging this should help unblock others to see about upgrading the other components.

ddohler (approved):

PR #464, opened 2017-10-18
Add geoserver service to dev environment

jisantuc:
Overview
-----

This PR creates a geoserver docker container based on a template from another project,
adds that service to docker-compose, and changes the default `MAP_SERVER` location
in the front-end code to point to this service.

Testing
-----

### Path 1

*Note*: you can forward 9091 to outside the VM and skip path 2 if you want

- from inside the vm, `./scripts/server`
- everything should come up without issue
- kill the server
- `./scripts/console django bash`
- `apt install wget`
- `wget geoserver.internal.districtbuilder.com:8080/geoserver/web/`
- you should download an `index.html` file

### Path 2

- from outside the vm, `./scripts/server` (make sure the vm is shut down)
- navigate a browser to `localhost:9091/geoserver/web/`
- check out your totally sweet geoserver landing page


geoserver/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# A GeoServer installation
+FROM java:8-alpine
```
hectcastro: The `java` container image in deprecated in favor of `openjdk`. In this case, I think we just want the JRE vs. JDK, so would be good to replace this with `openjdk:8-jre-alpine`.

geoserver/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+ENV GEOSERVER_DIR /usr/share
+ENV GEOSERVER_EXT_DIR $GEOSERVER_DIR/geoserver-$GEOSERVER_VERSION/webapps/geoserver/WEB-INF/lib
+ENV PATH /usr/share/geoserver-$GEOSERVER_VERSION/bin:$PATH
+ENV JAVA_HOME /usr/lib/jvm/java-1.8-openjdk
+ENV GEOSERVER_HOME $GEOSERVER_DIR/geoserver-$GEOSERVER_VERSION/
+
+RUN apk add --update ca-certificates openssl && update-ca-certificates && \
+  wget -O $GEOSERVER_DIR/geoserver.zip $GEOSERVER_ZIP_URL && \
+  unzip -qq $GEOSERVER_DIR/geoserver.zip -d $GEOSERVER_DIR
```
hectcastro: Not sure how big the ZIP file is, but we can likely remove it here as the last step.

geoserver/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+COPY data_dir /data
```
hectcastro: What is expected to make this directory available? It doesn't seem to be here when I get to this step in `update`.

geoserver/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# A GeoServer installation
+FROM java:8-alpine
+
+MAINTAINER Azavea
+
+ENV GEOSERVER_VERSION 2.11.2
+ENV GEOSERVER_ZIP_URL http://sourceforge.net/projects/geoserver/files/GeoServer/${GEOSERVER_VERSION}/geoserver-${GEOSERVER_VERSION}-bin.zip
```
hectcastro: Files hosted on SF could lead to issues later. I think we should leave things as is for now, but keep an eye on failures/speed to determine if we should put this up somewhere.

geoserver/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# A GeoServer installation
+FROM java:8-alpine
+
+MAINTAINER Azavea
```
hectcastro: This has been deprecated. We can get just gut it.

django/publicmapping/static/js/mapping.js
```diff
@@ -137,7 +137,7 @@ function init() {
 
     // default map_server is on same host unless otherwise specified 
     if (MAP_SERVER === "") {
-        MAP_SERVER = window.location.host;
+        MAP_SERVER = "geoserver.internal.districtbuilder.com:8080"
```
hectcastro: This wouldn't resolve outside of the VM, right?

jisantuc: I was under the impression that, running in the container, it would have access to the container's networking. Is that not the case?

geoserver/Dockerfile
jisantuc: I _think_ it's going to go away. I misread intent for this directory. Geoserver needs to know where to put its data, but we don't need to see it or care about it I think. I was under the impression originally that we'd be placing thing there directly.

notthatbreezy: This should probably be initialized with some data to start with that gets Geoserver in a running state. In [development](https://github.com/PublicMapping/DistrictBuilder/pull/464/files#diff-4e5e90c6228fd48698d074241c2ba760R45) those files get mounted directly. It might make sense to include some of those files in this PR if I understand correctly.

django/publicmapping/static/js/mapping.js
hectcastro: But isn't `MAP_SERVER` being resolved in the browser? This seems similar to how we handle `TILE_SERVER_LOCATION ` inside of the RF UI.

jisantuc: ...yeah. Ok yeah :man_facepalming: 

i think i thought something like:

"it's in django so it's on the server this is fine"

:man_facepalming:

hectcastro (approved):

PR #477, opened 2017-12-13
Make app more usable

jisantuc:
Overview
-----

This commit includes many usability changes. Summarizing everything would take a
while, so here's what it enables you to do:

Testing
-----

All `./manage.py` mentions refer to running inside a running django container

- copy `settings.py.in` to `settings.py` outside your vm
- `./scripts/update`
- `./scripts/server`
- `./scripts/load_development_data`
- also load the 0 geounit (`./manage.py setup with the `-g0` flag)
- `./manage.py setup -n0 config/config.xml`
- `./manage.py setup -t config/config.xml`
- `./manage.py collectstatic`
- enter the application as a guest
- view plans
- click on a plan
- load a page that should have a map, but doesn't because open layers
- go to pages where you'd be able to generate reports, if those endpoints worked


ddohler (approved):
I think this could use one more round of testing from a fresh provision; I haven't been able to get `./manage.py setup` to work because it complains about not being able to find the shapefile, even after having run `load_development_data` a couple times. I've also run into problems with some of the scripts leaving behind running containers that cause port conflicts with future containers. 

Having said all that, I was able to log into the app as a guest, so I think this is a giant step forward from where it was previously, and this achieves the goal of the mini-hackathon, which was to get the app into a usable-enough state to allow us to parallelize subsequent tasks.

jisantuc:
On that note, working through fixing geoserver setup, I'm encountering some other things that I think deserve some attention as part of provisioning fixups. The issue we have currently with the settings file getting overwritten a thousand times also happens if you try to make changes to the `config.dist.xml` file and you're not careful about where you make those changes. I'm creating a separate issue for templating out the config.dist.xml file using `group_vars`, unless there's a way to do environment references somehow from within an xml doc

pcaisse:
> I've also run into problems with some of the scripts leaving behind running containers that cause port conflicts with future containers.

For posterity, I've had the same issues. I've had better luck running `docker-compose build` and `docker-compose up` directly instead of using the scripts.

PR #478, opened 2017-12-21
Make geoserver and geolevel nesting setup flags work

jisantuc:
Overview
-----

This PR with an unreasonably large diff accomplishes a few things:

- updates geos interactions so that the `-nX` setup.py flags work
- updates RESTful interaction with geoserver so that geoserver configuration works
- updates settings, config.xml, and the django dockerfile to have a new place to store styles and to have accurate db info
- does a bunch of automatic styling cleanup

Notes
-----

You need to have geounits loaded for testing this PR. The easiest was to do that is with the `load_development_data` script. I included a commit that I'll rebase out that points to a copy of the shapefile with the lowest level polygons merged in my S3. Things will go faster if you use that one. If you do `load_development_data`, make sure you tell it to do all three geolevels in that import with `-g0 -g1 -g2`.

Also, if you opt for `./scripts/server` over `docker-compose up -d`, you'll see a _ton_ of geoserver errors flying by. Those are fine! They're 404s basically from checking to see if resources that don't exist already exist. It's ok, don't panic. (I panicked a lot)

Testing
-----

- do whatever incantations are required to get settings into your vm. I think this is adequate:
  - `vagrant up --provision`
  - `vagrant rsync-auto`
  - separately: `cp django/publicmapping/publicmapping/settings.py.in django/publicmapping/publicmapping/settings.py`
  - `vagrant reload`
  - `vagrant ssh`
  - confirm that the `SLD_ROOT` in `settings.py` is `/opt/sld/`
- `docker-compose build`
- `./scripts/server` (or `docker-compose up -d` if you don't want to be inundated with logs)

#### Confirm nesting flags work

- `docker-compose exec django bash`
- run the nesting setup flag (assuming you already have geounits loaded, otherwise, also load geounits, see the first note): `./manage.py setup -n0 -n1 -n2 config/config.xml`
- it should take some time and then succeed

#### Confirm the geoserver flag works

This is where things get fun (you didn't think they were fun yet, right?)

Creating style files depends on [`django-sld`](https://github.com/azavea/django-sld). `django-sld` is very old and I can't verify that it's not in use anywhere else, so I'm hesitant to change it yet. It uses an API for getting field information that was marked for deprecation in 1.10 and removed in 1.11. So! On to testing

- `docker-compose exec django bash` (unless you're still in there)
- install `vi` or `nano` or whatever you want to use
- edit `/usr/local/lib/python2.7/site-packages/djsld/generators.py` to replae `get_field_by_name(geofield)[0]` with `get_field(geofield)` and save it
- `./manage.py setup -G config/config.xml`
- navigate to `localhost:9091/geoserver/web/` in a browser and look at all the THINGS you made (layers, stores, styles...)


django/publicmapping/redistricting/config.py
```diff
@@ -1367,13 +1677,18 @@ def _rest_config(self, method, url, data=None, headers=None):
             conn = httplib.HTTPConnection(self.host, self.port)
             conn.request(method, url, data, headers)
             rsp = conn.getresponse()
-            rsp.read() # and discard
-            conn.close()
+            rsp.read()  # and discard
+            conn.close(
```
pcaisse, line 1681: :thinking:

pcaisse:
I'm having an issue here:
```
root@4c48da496d71:/usr/src/app# ./manage.py setup -G config/config.xml
Geoserver configuration cleaned.
Could not create Namespace
root@4c48da496d71:/usr/src/app# ./manage.py setup -G config/config.xml
Exception AttributeError: "'_MOFileParser' object has no attribute 'fhandle'" in <bound method _MOFileParser.__del__ of <polib._MOFileParser object at 0x7fa4052b1790>> ignored
Geoserver configuration cleaned.
Could not create Datastore
root@4c48da496d71:/usr/src/app# ./manage.py setup -G config/config.xml
Geoserver configuration cleaned.
Could not create Datastore
root@4c48da496d71:/usr/src/app# ./manage.py setup -G config/config.xml
Exception AttributeError: "'_MOFileParser' object has no attribute 'fhandle'" in <bound method _MOFileParser.__del__ of <polib._MOFileParser object at 0x7f1457168350>> ignored
Geoserver configuration cleaned.
Could not create Datastore
```
I'm getting different results when I run the command multiple times. I first couldn't create the namespace and then I couldn't create the datastore. Also, sometimes that exception is raised and sometimes it isn't.

I see that the response in `_rest_config` is returning a 404 each time but I'm not sure what to do about it. 

I took a look at `django/publicmapping/redistricting/config.py` to further diagnose the issue but the refactoring made it hard to tell where the actual changes were.

pcaisse:
Okay, I did nothing different but it worked!

```
root@94154ec5ce5a:/usr/src/app# ./manage.py setup -G config/config.xml
Geoserver configuration cleaned.
Geoserver configuration complete.
```

After having checked out another branch, I realized all the geoserver logging was being written to stdout so I checked this branch out again and ran `docker-compose up` in the foreground, hoping to get some insight into what was wrong with geoserver, and it just worked. ¯\\\_(ツ)\_/¯

pcaisse:
What is the deal with manually editing `/usr/local/lib/python2.7/site-packages/djsld/generator.py` Is that a versioning thing? Is it broken?

django/publicmapping/redistricting/config.py
```diff
@@ -1288,20 +1563,24 @@ def _check_spatial_resource(self, url, name, dictionary, update=False):
         @keyword update: Optional. Update the featuretype if it exists?
         @returns: A flag indicating if the configuration call completed successfully.
         """
+        # TODO: this throws a 404 from geoserver every time it can't find something,
+        # and we expect it not to be able to find something pretty often. While I wouldn't
```
pcaisse, line 1567: What does this mean exactly? Are these geoserver endpoints just flakey?

kshepard:
Just to add another data point, I also had the issue with: `Could not create [Namespace|Datastore]`. It only went away after I restarted my VM and tried again.

django/publicmapping/redistricting/config.py
jisantuc: No, it's by design -- `_rest_check` attempts to read a resource that, if it existed, would have a known name and the geoserver routes use the resource names. If it's there, there's logic to avoid trying to create it. If it's not there, it creates it. It discovers that it's not there through the 404 though. It's not that it's broken, just that there's a really low signal-to-noise ratio in the geoserver container's stdout.

jisantuc:
> What is the deal with manually editing /usr/local/lib/python2.7/site-packages/djsld/generator.py Is that a versioning thing? Is it broken?

It uses an old API. [We own the repo](https://github.com/azavea/django-sld) and it's probably the case that nothing else uses it and that we could safely PR a fix to use the field access API from a non-deprecated django, but I was hesitant to do that because I don't really know what the surface area of "fix deprecated django patterns" is.

> I checked this branch out again and ran `docker-compose up` in the foreground, hoping to get some insight into what was wrong with geoserver

You can also `docker-compose logs -f containerName` if containers are running in the background and you want to see what they're up to.

jisantuc:
That `Could not create DataStore` bug is really weird -- `Geoserver configuration cleaned.` is what it tells you after deleting everything from geoserver. I have no idea why it'd be unable to create things into a blank slate. :thinking:

pcaisse:
> we could safely PR a fix to use the field access API from a non-deprecated django

I think it would make sense to cut a new release, perhaps a new major revision to show that we now a require a new-ish version of Django.

jisantuc:
I was about to say "I'll open an Asana task" but then I checked and found out I already made one. I agree. My short-term concern in estimating that is that I've been unsuccessful running tests locally in `django-sld` and there's no CI, so it'd be a little bit difficult to prove correctness before cutting a new release.

pcaisse (changes requested):
Just to reiterate from the Slack thread, before merging I'd like to:
* Bring the `generator.py` function into DistrictBuilder and get rid of `django-sld` as a dependency
* Update README with how to run setup flags

One other thing that would be really helpful for reviewing is breaking out any reformatting/style changes into a separate commit. I can't really tell what's going on in `django/publicmapping/redistricting/tasks.py` because the diff is so huge.

pcaisse (approved):

pcaisse:
I created a story to update documentation here: https://www.pivotaltracker.com/story/show/154063759

jisantuc:
I think that's still a relevant, though I did end up capturing some of that here because I didn't see your comment beforehand.

pcaisse:
I don't really understand the utility of 9a10dd051202c4cfd36c314bc485bb9f51ee2013. As it stands now I have to remember to run `./scripts/reconfig` when I switch branches, whereas before I didn't have to do that.

PR #480, opened 2018-01-05
Fix OpenLayers

ddohler:
# Overview
This fixes the "Invalid array length" error thrown by OpenLayers when trying to load the plan view page. The issue appears to have been a change in the behavior of GeoDjango although I didn't go through the Django changelogs to verify. The `MultiPolygon.coords` attribute currently returns the coordinates for the entire geometry, but the front-end Javascript is expecting a 4-element bounding box array of [xmin, ymin, xmax, ymax]. Additionally, the coordinates were returned as a tuple rather than an array, which is evaluated by Javascript as the final element of the tuple, i.e. (w, x, y, z) => z. This fixes both problems.

This also fixes a problem with `MEDIA_ROOT` being used to access `.po` files; Django has since introduced a `STATIC_ROOT` variable that should be used in this case (and indeed, we are currently storing the `.po` files in `STATIC_ROOT`, rather than `MEDIA_ROOT`), so this switches to that.

# Demo
screen shot 2018-01-05 at 11 11 19 am

# Testing instructions
- Get your VM and `./scripts/server` started. This assumes you've run all `load_development_data` and `./manage.py setup` commands necessary to get a basic instance going.
- Run `docker exec -ti vagrant_django_1 bash` and then `./manage.py collectstatic --noinput` to make sure you have updated static files (this needs to happen every time rsync triggers so if you run into errors about OpenLayers not being defined, just rerun it).
- Log in as guest and view the example plan.
- You should see a map, and it should be centered on Virginia. 🎉

### DIFF
--- a/django/publicmapping/redistricting/models.py
+++ b/django/publicmapping/redistricting/models.py
@@ -82,11 +82,11 @@ class BaseModel(models.Model):
         lang = translation.get_language()
         if not lang in I18N_CACHE:
             try:
-                path = os.path.join(settings.MEDIA_ROOT, '../locale/%s/LC_MESSAGES/xmlconfig.mo' % lang)
+                path = os.path.join(settings.STATIC_ROOT, '../locale/%s/LC_MESSAGES/xmlconfig.mo' % lang)
                 path = os.path.normpath(path)
                 I18N_CACHE[lang] = polib.mofile(path)
             except Exception, ex:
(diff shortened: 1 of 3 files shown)


jisantuc (approved):
:tada:

For extra credit I zoomed in to different levels and watched the failing geoserver requests and confirmed that they're still routed to county/vtd/block reasonably.

PR #481, opened 2018-01-08
Fix reporting

pcaisse:
## Overview
This PR makes reporting work. It punts on all major architectural decisions just to get things working.

## Testing
1. Run `./scripts/reconfig`
1. `vagrant up && vagrant ssh`
1. Within VM, `docker-compose build && docker-compose up`
1. Log in to DistrictBuilder as a guest
1. Go to the "Evaluate" tab (http://localhost:8080/districtmapping/plan/1/view/)
1. Click any report and then click "Create and Preview New Report"
1. See the report in the preview. Notice that the "Open Current Report in New Window" also works.

## Notes
This PR does not address much tech debt that carried over from the original DistrictBuilder. It does pretty much the bare minimum to get things working. Here is the outstanding tech debt that is carrying over as I see it:
* Continues to use [django-celery](https://github.com/celery/django-celery) which is old, looks abandoned, and which we don't need to use anymore AFAICT
* Uses Django as the message broker for celery which is [not recommended](http://docs.celeryproject.org/en/3.1/getting-started/brokers/django.html)
* Upgraded Celery to `3.1.25` but cannot upgrade to the newest version `4.1.0` because that version is not supported by `django-celery` (I found this out the hard way because pip isn't great at dependency resolution...)
* Uses `C_FORCE_ROOT=1` to run Celery as root, which the 4.1.0 documentation calls ["a very dangerous practice"](http://docs.celeryproject.org/en/latest/userguide/daemonizing.html#running-the-worker-with-superuser-privileges-root)
* Continues storing the fully qualified name of the calculator module in the database (predictably, this turned out not to be a robust solution):
```
In [1]: from redistricting.models import ScoreFunction

In [2]: sf = ScoreFunction.objects.all()[0]

In [3]: sf
Out[3]: <ScoreFunction: a_congressional_population label>

In [4]: sf.calculator
Out[4]: u'publicmapping.redistricting.calculators.Interval'
```
* Does not resolve the issue of serving files we generate, such as reports. Previously they were served by Apache and we could use Nginx to do the same thing but that was too much to tackle at the moment.

However, it does make things work and upgrades our deps as far we can.

We can talk about how to address all of the above issues.

### DIFF
--- a/django/publicmapping/Dockerfile
+++ b/django/publicmapping/Dockerfile
@@ -13,4 +13,4 @@ COPY . /usr/src/app
 
 WORKDIR /usr/src/app
 
-ENTRYPOINT ["/usr/local/bin/gunicorn"]
+ENTRYPOINT ["/usr/src/app/run_django.sh"]
--- a/django/publicmapping/publicmapping/settings.py.in
+++ b/django/publicmapping/publicmapping/settings.py.in
@@ -29,6 +29,14 @@ DEBUG = True
 ALLOWED_HOSTS = []
 
 
+# Settings for django-celery process queue
+import djcelery
+djcelery.setup_loader()
+
+# TODO: Don't use Django as broker
+BROKER_URL = 'django://'
+
+
 # Application definition
 
 INSTALLED_APPS = [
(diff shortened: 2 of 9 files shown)


pcaisse:
One thing to note is that 8b2561787b694a965f3c6fae0c94fc51edec5383 can be cleaned up significantly with a small change to the nginx config once https://github.com/PublicMapping/DistrictBuilder/pull/486 is merged.

PR #482, opened 2018-01-09
Don't require manage.py collectstatic on every server run

pcaisse:
## Overview
Fixes issue where `collectstatic` would have to be run each time a file was changed.

## Testing
`docker-compose build && docker-compose up` should do it. Notice you don't need to run `collectstatic` to see the map (kind of) working before or after any changes.

## PivotalTracker
https://www.pivotaltracker.com/n/projects/2140015/stories/154063082

## Notes
I've uncovered some weirdness around how we serve static files. Currently we have two `static/` directories checked into source control, one of which is our `STATIC_ROOT`:
```
10:29 $ find -type d -name static
./django/publicmapping/redistricting/static
./django/publicmapping/static
```
They both contain some of the same files, some of which are only slightly different. Presumably the common files in our `STATIC_ROOT` (`django/publicmapping/static`) are getting overwritten by the app-specific static files when we run `collectstatic`, though I'm not entirely sure of that.

@kshepard has mentioned doing away with `collectstatic` entirely which might be a good option but I'm going to hold off on any major changes for now just so we can move forward and address this immediate annoying issue.

It's also worth noting that with our current setup, if you're doing frontend work you would still need to run `collectstatic` to have your changes to static files reflected, which is something that we should address.

### DIFF
--- a/django/publicmapping/Dockerfile
+++ b/django/publicmapping/Dockerfile
@@ -13,4 +13,6 @@ COPY . /usr/src/app
 
 WORKDIR /usr/src/app
 
+RUN python manage.py collectstatic --noinput
+
 ENTRYPOINT ["/usr/local/bin/gunicorn"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,7 @@ services:
       - "8081:8081"
     volumes:
       - ./django/publicmapping/:/usr/src/app/
+      - /usr/src/app/static
     command:
       - "--workers=2"
       - "--timeout=60"


ddohler:
One issue I had was that I needed to run collectstatic every time I made a file change, not just on initial server start -- do you see the same behavior, and does this address that?

pcaisse:
@ddohler Would you mind testing out this PR to see if you still have that issue?

It seems likely that you would have to run `./manage.py collectstatic` after changing static files. My understanding is that if you serve static files in dev using `runserver` then it won't collect the static files for you and they'll just be served from their original location and you don't have to worry about running `collectstatic`, but we're not doing that currently.

The intent here is to avoid having to do `docker-compose exec django bash -c "./manage.py collectstatic"` after every `docker-compose up`.

ddohler:
I'm still seeing the issue even after checking out this branch. Can you reproduce following these steps?
1. Check out this branch, `vagrant up` and have `vagrant rsync-auto` running
2. `docker-compose build && docker-compose up` from inside the VM
3. Log in as guest and then go to view `http://localhost:8080/districtmapping/plan/1/view/`
4. It should work, in the sense that the map should appear and tiles should load
5. Open `django/publicmapping/redstricting/views.py` and make an inconsequential change.
6. Wait for `rsync-auto` to finish
7. Reload the page and the map should disappear; there should be a console error saying `OpenLayers is not defined`
8. Running `collectstatic` manually as above should solve the problem until the next time a file is changed.

jisantuc:
I think this should be in `scripts/server` -- https://github.com/github/scripts-to-rule-them-all#scriptserver

pcaisse:
Hmm.... yeah, you're absolutely right. I can reproduce it. Not sure how I wasn't encountering this in development up until now.

pcaisse:
@ddohler ae264cb75d2c2fde71fcc79720f64b6e01cd7685 should fix the issue

pcaisse:
@jisantuc I'm not sure I understand how I can put the `manage.py` commands inside of `script/server`. Doesn't the container need to be running to run migrations, collect static, start Gunicorn, etc? I was thinking of the entry point script (`run_django.sh`) as what happens when you run `docker-compose up`. Let me know if I'm thinking about it wrong.

ddohler:
Yep, that fixed it for me! I agree with @jisantuc that the container entrypoint isn't quite the right place to put this. If the problem was more about a missing docker volume and this is now a one-time task, then I think this should be able to live in `scripts/setup`, perhaps?

jisantuc:
I misunderstood, feel free to disregard my previous comment.

django/publicmapping/run_django.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+python manage.py migrate
+python manage.py collectstatic --noinput
```
hectcastro: This command could probably be moved into the `Dockerfile` after a step that copies in the source code. Changes to the source code in a copy step would invalidate directives below it, which would allow `collectstatic` to become aware of any new static files.

See: https://github.com/azavea/collab-for-children/blob/develop/src/django/Dockerfile#L16-L19

django/publicmapping/run_django.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+python manage.py migrate
```
hectcastro: I don't think this is the type of command we want to be running when the container is instantiated. Usually, we try to break out management commands that require database interaction such that they're run after deployments. For example, if we wanted to achieve zero-downtime deployment, and we already had a version of the application running (v1)—the instantiation of a newer version of the same service (v2) could mutate the database in a way that breaks v1 (e.g., dropping a column) before v2 ever becomes health enough to do work.

In this case, I think that moving the invocation of `manage.py migrate` to `scripts/update` would be better.

django/publicmapping/run_django.sh
pcaisse: @hectcastro How does this work for frontend development? Do you have to run `collectstatic` or rebuild whenever you change a static file?

hectcastro: I have not tested this, but the static assets `collectstatic` manages should be served up in development when `DEBUG=True` if you do something like:

```python
from django.conf import settings
from django.conf.urls.static import static

urlpatterns = [
    # ... the rest of your URLconf goes here ...
] + static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
```

See: https://docs.djangoproject.com/en/1.11/howto/static-files/#serving-static-files-during-development

PR #483, opened 2018-01-10
Python style cleanup

shreshthkhilani:
## Overview
Add a [yapf](https://github.com/google/yapf) pre-commit hook using the [pre-commit framework](http://pre-commit.com/). Add setup for this framework in `scripts/setup`. 

Also clean-up existing python code with `yapf`.

## Testing
Check to see that the branch isn't broken. Check if you can enter as guest.

## Notes
To use the pre-commit hooks, we need to rerun the `setup` script on our local environments. This also means that anyone setting up the repo needs to have Python and pip installed.


pcaisse:
I'm not sure how I feel about git commit hooks being an official part of our development workflow. I like them fine when I set them up but I don't love being forced into using one. I'd prefer to have a linter set up that would break the build (once we have one of those ;)). That said, it wouldn't be the end of the world for me either.

shreshthkhilani:
I agree that I don't feel wonderful about forcing pre-commit hooks and I'd prefer the lint file too. However, in this case, the linter would return anything between 300 and 800 violations with the existing code and so we'd have to fix those first.

scripts/setup
```diff
@@ -22,6 +22,10 @@ then
         # Ensure that a directory exists to house AWS profiles.
         mkdir -p "${HOME}/.aws"
 
+        # Sets up pre-commit hooks to yapf new python files.
+        pip install pre-commit
```
jisantuc: I think this version should be pinned. Someone provisioning for the first time who already has `pre-commit` installed but an old version would have no changes here, and I don't know whether we're certain about version compatibility.

Also, it should probably move into an ansible role as a `local_action`, so that we can turn it on and off depending on whether it's a development setup or not (see, e.g., https://github.com/azavea/collab-for-children/tree/develop/deployment/ansible/roles/collab-for-children.shellcheck/tasks).

jisantuc (changes requested):
I'm still bringing up my vm after yesterday's series of small catastrophes, but I'm expecting that to go fine. Just the version pin for `pre-commit` needs to change.

shreshthkhilani:
@jisantuc successfully added the pre-commit install to a local_action ansible task.

jisantuc (approved):
`./scripts/setup` as single entrypoint :tada:

PR #485, opened 2018-01-10
Revert "Revert "Fix reporting""

pcaisse:
This reverts commit 95ec4e5ef93229576a83b0ebd7713a568e6ceab7.

## Overview

I accidentally merged https://github.com/PublicMapping/DistrictBuilder/pull/481 :shame: and then had to revert. This reverts that reversion.

Please refer to that PR for a nice PR descriptions and nice commits.

### DIFF
--- a/django/publicmapping/run_django.sh
+++ b/django/publicmapping/run_django.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+# Run celery worker in background
+# TODO: Don't use C_FORCE_ROOT
+C_FORCE_ROOT=1 python manage.py celery worker --loglevel=info &
+
+# Run Django with Gunicorn
+/usr/local/bin/gunicorn \
+  --workers=2 \
+  --timeout=60 \
+  --bind=0.0.0.0:8080 \
+  --reload \
+  --log-level=debug \
+  --access-logfile=- \
+  --error-logfile=- \
+  -kgevent \
+  publicmapping.wsgi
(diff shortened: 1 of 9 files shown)


jisantuc:
Can you assign this elsewhere? I've spent 2 hours at war with virtualbox and had to kill all of my VMs again :rage4:

django/publicmapping/run_django.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+# Run celery worker in background
+# TODO: Don't use C_FORCE_ROOT
+C_FORCE_ROOT=1 python manage.py celery worker --loglevel=info &
```
ddohler, line 5: Running multiple processes in a single Docker container is generally [considered a no-no](https://docs.docker.com/engine/admin/multi-service_container/). In this case we could set up a `celery` service that uses the `image` parameter to ensure that it works off the same image as the `django` service.

pcaisse: Yeah, it's bad, but so many things about this PR are bad.

I think that to have a standalone celery service we'd need to upgrade celery and its API has breaking changes. That would also necessitate doing away with `django-celery` which doesn't support new celery.

I'm not sure what to do with this PR at this point. It almost feels like a spike in that now we know everything that's involved in getting it working.

Do you think the best course of action would be to let it serve as a reference and just open up a new PR with "reporting done right"?

ddohler:
Surfacing a verbal conversation, we're going to hold on this for now because so much of it is hacky due to Celery not being upgraded yet. Upgrading Celery is a task for this sprint, so once that's finished, we will revisit this and cherry-pick over any commits that are not Celery-related.

PR #486, opened 2018-01-11
Use nginx to serve static files

pcaisse:
## Overview
This PR sets up a container for nginx in front of Gunicorn/Django. nginx is now serving static files instead of Gunicorn, which means that we no longer have to run `collectstatic` and for frontend development we can simply make changes to static files and see them reflected upon refresh.

I'd recommend looking at 1ade1d5 by itself first since the second commit has a ton of deleted files in it.

## Testing Instructions

 * Run `./scripts/reconfig` (settings file has changed)
 * `vagrant up && vagrant ssh` into VM
 * Within the VM, run `docker-compose build && docker-compose up`
 * Go to http://localhost:8080/districtmapping/plan/1/view/
 * Notice everything looks the same as before in the browser, but that nginx is serving static assets and forwarding all other requests:

```
django_1     | 172.18.0.6 - - [11/Jan/2018:16:50:56 +0000] "POST /districtmapping/plan/1/demographics/ HTTP/1.0" 500 985 "http://localhost:8080/districtmapping/plan/1/view/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36"
nginx_1      | 10.0.2.2 - - [11/Jan/2018:16:50:56 +0000] "POST /districtmapping/plan/1/demographics/ HTTP/1.1" 500 985 "http://localhost:8080/districtmapping/plan/1/view/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36" "-"
django_1     | [2018-01-11 16:50:56 +0000] [9] [DEBUG] Closing connection.
nginx_1      | 10.0.2.2 - - [11/Jan/2018:16:50:57 +0000] "GET /static/openlayers/theme/default/style.css HTTP/1.1" 200 9952 "http://localhost:8080/districtmapping/plan/1/view/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36" "-"
nginx_1      | 10.0.2.2 - - [11/Jan/2018:16:50:57 +0000] "GET /static/openlayers/img/zoombar.png HTTP/1.1" 200 745 "http://localhost:8080/districtmapping/plan/1/view/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36" "-"
```
 * Make some trivial change to a static asset and notice that changes are reflected without running `collectstatic`.
 * For extra credit, within the VM, do the following to verify that the collected static files (`admin` and `django-extensions`) folders are in the `nginx` container:

```
vagrant@vagrant-ubuntu-trusty-64:/vagrant$ docker-compose exec nginx bash
root@dcf92b5aae75:/# ls -al static/
total 40
drwxr-xr-x 10 root root 4096 Jan 11 18:55 .
drwxr-xr-x 37 root root 4096 Jan 11 18:58 ..
drwxr-xr-x  6 root root 4096 Jan 11 18:54 admin
drwxr-xr-x  2 root root 4096 Jan 11 18:34 css
drwxr-xr-x  5 root root 4096 Jan 11 18:54 django_extensions
drwxr-xr-x  2 root root 4096 Jan 10 14:45 images
drwxr-xr-x  5 root root 4096 Jan 11 18:34 jqGrid
drwxr-xr-x  6 root root 4096 Jan 10 14:45 jquery
drwxr-xr-x  2 root root 4096 Jan 11 18:34 js
drwxr-xr-x  4 root root 4096 Jan 11 18:34 openlayers
```

## Notes
~While nginx is now serving static assets and we don't have to run `collectstatic` for frontend development to see changes, we do still need to run `collectstatic` in `Dockerfile`. This is because we're still using static assets from some Django and 3rd party installed apps.~ EDIT: This is no longer true. I've committed the only collected static files we were using (see 5f26035). We shouldn't need to think about `collectstatic` for the forseeable future.

As an example of those static assets actually being used somewhere, see `admin/css/forms.css` collected from `django.contrib.admin` in: https://github.com/PublicMapping/DistrictBuilder/blob/develop/django/publicmapping/publicmapping/templates/admin/redistricting/subject/upload_form.html#L101)


kshepard (approved):
This looks good to me after pinning the nginx version. I was able to run through all of the tests successfully, except for one, which I'm thinking may be more a problem with my local setup. I wasn't able to edit a static file and see the changes propagate to the container/browser for some reason (makes it to the VM though).

PR #487, opened 2018-01-15
Route to geoserver via nginx

jisantuc:
## Overview

This PR routes all traffic to geoserver through nginx and makes generated
`sld` files available to the containers that need them.

### Checklist

### Demo

image

### Notes

I know, I know -- "There may be other problems with Geoserver at this point, which will be covered by other tasks." -- geoserver not being able to find `sld` files was affecting my ability to tell whether I had configured nginx correctly or not.

Also,  `django` expects to get `sld`s from an endpoint, and geoserver expects them to be in a specific location in the geoserver data directory. Moreover, django generates the files and puts them somewhere in the django container, despite not caring about their final destination. This is convoluted, but I think it's our fate.

## Testing Instructions

It's possible this will fail and you'll need to destroy, recreate, and reconfigure your `geoserver` container. It's also possible that was a result of some sort of local artifact. I don't know. I hate 💻s now. 

 * `mkdir sld` from the project root
 * `docker-compose down`, `docker-compose up`
 * navigate to `:8080/geoserver/<whatever>` and verify that you don't get stopped by django's router not knowing what's going on
 * navigate through the main app and verify that I didn't appear to kill any other routing configuration


nginx/django.conf
```diff
@@ -2,15 +2,27 @@ upstream web {
   server django:8080;
 }
 
+upstream geoserver {
+    server geoserver:8080;
```
pcaisse: Looks like we have mixed spacing going on here. Can we stick with 2 spaces?

pcaisse (changes requested):
LGTM. One small nitpick and you're good.

FYI, I don't think you need to create the `sld` directory manually. It got created when I created the geoserver layers via `./manage.py setup -G`.

nginx/django.conf
jisantuc: Sure, I'll change my .spacemacs config for nginx mode so I don't mess that up again in the future (otherwise I 100% will)

pcaisse (approved):

PR #488, opened 2018-01-15
Fix usage of static template tag so that jqGrid loads

ddohler:
## Overview

jqGrid wasn't being loaded because we were attempting to do a variable interpolation within a template tag, which isn't allowed. I'm not a huge fan of the fix but it seems to be standard practice. https://stackoverflow.com/questions/16655851/django-1-5-how-to-use-variables-inside-static-tag

### Checklist

### Demo

Take a gander at these beautiful paginator arrows:
dbjqgrid

## Testing Instructions

 * Enter as a guest. 
 * You should see pagination controls on the plan list view, and they shouldn't have a big `undefined` obscuring them.

[fix #154033194]


jisantuc (approved):
image

stylish 😎

PR #489, opened 2018-01-15
Make the existing tests run using scripts/test

shreshthkhilani:
## Overview

This PR makes the existing `redistricting/tests.py` run. It fixes enough Django and database problems to make the tests run, but doesn't fix the errors in the tests themselves. 

This also calls changes the STRTA `script/test` to run these tests to make it easier to set up CI. 

### Checklist

### Notes

The test db that Django stands up already has a row in the `django_site` table with "example.com" as the domain. Since this was causing a collision with the fixtures file, the fixtures file domain was changed to "examples.com". 

 A Django 1.8 issue made hardcoding the language necessary for testing. Refer to [this issue](https://github.com/django-parler/django-parler/issues/90).

## Testing Instructions

 * `vagrant ssh`
 * `./scripts/test`
 * Confirm that the tests run (not that they pass; they don't).


jisantuc (changes requested):
One thing that would be helpful would be if the test classes split into separate files in a `tests/` directory, which I think is more like "best practice" at this point. If nothing else it will make our lives a lot better when we start working on fixing the failing tests.

shreshthkhilani:
@jisantuc broken up into multiples files in the /tests dir

django/publicmapping/tests/test_config.py
```diff
@@ -0,0 +1,140 @@
+from django.test import TestCase
+import tempfile
+import os
+import logging
+# from redistricting.tasks import *
```
jisantuc: Remove these stray commented imports

django/publicmapping/tests/test_multi_member.py
```diff
@@ -0,0 +1,429 @@
+from base import BaseTestCase
+
+import os
+import zipfile
+from django.test.client import Client
+from redistricting.models import Geolevel, Geounit, District, Plan, Subject
+from redistricting.tasks import DistrictIndexFile
+from redistricting.calculators import (Range, Equipopulation, Equivalence,
+                                       Interval, MultiMember)
+# from redistricting.reportcalculators import *
```
jisantuc: Remove

django/publicmapping/tests/test_plan.py
```diff
@@ -0,0 +1,1318 @@
+from base import BaseTestCase
+
+from django.db.models import Sum, Min, Max
+from django.test.client import Client
+from django.contrib.auth.models import User
+
+from redistricting.models import *
```
jisantuc: Probably open a story in PT for removing all remaining `import *`s

django/publicmapping/tests/test_plan.py
```diff
@@ -0,0 +1,1318 @@
...
+            District.objects.filter(
+                plan=self.plan, district_id=self.district2.district_id + 1),
+            key=lambda d: d.version)
+        district3units = district3.get_base_geounits(0.1)
+
+        self.assertEqual(486, len(district3units),
+                         'Incorrect number of geounits returned in dist3: %d' %
+                         len(district3units))
+
+        # Plan looks like this now:
```
jisantuc: I had no idea we had ascii art in this project, I love this.

jisantuc (changes requested):
Omg so much more readable

Can you move the `tests/` dir into the redistricting app (since they're tests of the `redistricting` app's models)? Other than that, just a little bit of cleanup. This is great.

shreshthkhilani:
Story created, directory moved, and comments removed. Thanks!

jisantuc (approved):
I'd say "pending Travis," but we both know how well that's going to go. :+1:

PR #491, opened 2018-01-16
Check in settings.py

pcaisse:
## Overview

Check `settings.py` in so we don't need to worry about copying it over when settings change.

### Notes

Not doing anything with `config.xml` for now.

## Testing Instructions

* `docker-compose build && docker-compose up`
* Run `vagrant rsync-auto` separately if not already running

The app should come up normally and any changes to `settings.py` should show up in the VM and in source control.


ddohler (approved):
Works great. Changing the log level to `DEBUG` showed up in source control and also triggered an auto rsync followed shortly thereafter by a django worker reload.

PR #492, opened 2018-01-18
Fix reporting (part deux)

pcaisse:
## Overview

Fix reporting by upgrading Celery and creating a separate `celery` service.

### Notes

This PR cherry-picks some commits from https://github.com/PublicMapping/DistrictBuilder/pull/481 which was a much hackier way of implementing this same functionality.

## Testing Instructions

1. `vagrant up && vagrant ssh`
1. Within VM, `docker-compose build && docker-compose up`
1. Log in to DistrictBuilder as a guest
1. Go to the "Evaluate" tab (http://localhost:8080/districtmapping/plan/1/view/)
1. Click any report and then click "Create and Preview New Report"
1. See the report in the preview. Notice that the "Open Current Report in New Window" also works.

### DIFF
--- a/django/publicmapping/Dockerfile
+++ b/django/publicmapping/Dockerfile
@@ -6,6 +6,8 @@ RUN mkdir -p /opt/sld
 RUN mkdir -p /opt/reports
 
 # Create reporter user that can write to reports directory
+# This user is used to run `celery worker` for reporting
+# as it is not recommended to run Celery services as root.
 RUN useradd -m reporter
 RUN chown reporter: /opt/reports
 
--- a/django/publicmapping/Dockerfile
+++ b/django/publicmapping/Dockerfile
@@ -2,6 +2,13 @@ FROM quay.io/azavea/django:1.11-python2.7-slim
 
 RUN mkdir -p /usr/src/app
 RUN mkdir -p /opt/sld
+
+RUN mkdir -p /opt/reports
+
+# Create reporter user that can write to reports directory
+RUN useradd -m reporter
+RUN chown reporter: /opt/reports
+
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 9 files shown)


django/publicmapping/Dockerfile
```diff
@@ -2,6 +2,13 @@ FROM quay.io/azavea/django:1.11-python2.7-slim
 
 RUN mkdir -p /usr/src/app
 RUN mkdir -p /opt/sld
+
+RUN mkdir -p /opt/reports
+
+# Create reporter user that can write to reports directory
```
ddohler, line 8: This is to keep celery from complaining about being run as root, right? That might be worth a comment to explain because it's a bit unusual to see user permissions stuff inside a Dockerfile.

django/publicmapping/redistricting/tasks.py
```diff
@@ -20,19 +20,19 @@
...
 
-Author: 
+Author:
     Andrew Jennings, David Zwarg
 """
 
-from celery.task import task
-from celery.task.http import HttpDispatchTask
+from publicmapping.celery import app
+# from celery.task.http import HttpDispatchTask
```
ddohler: If there's no or a low chance that we'll re-enable this later, I'd just delete it.

ddohler (commented):
This looks good to me but I suspect I'll have trouble testing because my development data is all messed up right now. I'll give it a try tomorrow morning though.

django/publicmapping/Dockerfile
pcaisse: I added some more comments just after line 8

ddohler (approved):
Turns out that I was able to test after all; I saw the celery activity in the console and got back a result with 0 values.

PR #493, opened 2018-01-22
Fix login

pcaisse:
## Overview

Fix login. Previously logging in was redirecting to the admin login page and now it actually logs you in.

## Testing Instructions

 * `docker-compose build && docker-compose up` within VM
 * Create a user via "Sign Up" button
 * First try loging in with an incorrect password to check validation and notice the modal telling you authentication failed
 * Log in with correct username and password and notice it succeeds
 * Click "My Account" in upper right to see it is in fact the correct user
 * Click "Logout" and notice you are logged out correctly 

## PivotalTracker
https://www.pivotaltracker.com/story/show/154033115


jisantuc (approved):
worked considerably better after I checked out your branch 🤦‍

PR #494, opened 2018-01-24
Fix various problems with translations

kshepard:
## Overview

There were several problems with translations that are addressed here:
 * The language dropdown on the landing page was empty
 * The language dropdown after logging in listed every language known to Django
 * Translations were unable to load due to a few misconfigurations in settings.py
 * The `makelanguagefiles` management command had a bug where individual language codes were being split up and run for each character in the language code
 * There was a short-term workaround with `ENABLED_LANGUAGES` in a couple places, which has been removed

### Checklist

### Demo

district-builder-spanish

### Notes

 * There is a problem with preserving the compiled language files in the Docker filesystem. They are generated within the container via `makelanguagefiles`, and aren't persisted across restarts. I will add a task for this.
 * The languages are hardcoded to English and Spanish in the settings file. These need to be made dynamic as part of the task which implements the usage of a `.env` file.

## Testing Instructions

 * Provision (may need to comment out the pre-commit hook, which currently is problematic)
 * `vagrant ssh`
 * `scripts/setup`
 * `docker-compose exec django ./manage.py makelanguagefiles` -- this is required at the moment due to the problem discussed above in `Notes`
 * Browse to http://localhost:8080
 * Click the language picker dropdown. There should be options for `English` and `Spanish`.
 * Select `Spanish` and verify that the page reloads and is translated.


pcaisse (approved):
I feel strongly that we should've gone with Portuguese first, but I guess this is working :wink:

PR #495, opened 2018-01-25
Fix error being thrown when attempting to draw a plan

pcaisse:
## Overview

Fix error being thrown when attempting to draw a plan after logging in.

## Testing Instructions

Just need to check the branch out and `docker-compose up`. Follow the steps to reproduce here and make sure you don't see the error:
https://www.pivotaltracker.com/story/show/154607989


kshepard (approved):
Nice, glad this was straightforward. I verified that there are no other `render` calls fitting the invalid pattern.

PR #496, opened 2018-01-26
Speed up data loading (Part 1: Bulk Create)

ddohler:
## Overview

Switches `setup.py` to use Django's `bulk_create()` function to create Geounit `Characteristic`s.

Also switches a lot of the logging statements in `setup.py` to use `logger.info` rather than `logger.debug`; the script is usually being run semi-interactively so it seems like it should provide more feedback than it currently does.

### Checklist

### Demo

I instrumented the heck out of the code, and here's what the timings looked like. All numbers are aggregated over 500 Geounits. The numbers in parentheses are percentage of total. The projected times are probably low because I expect the existence query steps to start taking longer as more Geounits are inserted (but the query is indexed so it should be a logarithmic increase).

```

Original
--------------------------
Levels Queries 1.219147 (0.020924)
Shapefile attributes 0.662532 (0.011371)
Existence Query 1.398251 (0.023998)
Geometry 1.326095 (0.022759)
Save 5.832781 (0.100106)
Geounit Characteristics 47.707065 (0.818782)
Total 58.265918
Projected time: 58 / 500 * 285762 / 60 / 60 = 9.2 hours


Single get + batch create
--------------------------
Levels Queries 1.175547 (0.030687)
Shapefile attributes 0.646928 (0.016888)
Existence Query 1.351465 (0.035279)
Geometry 0.878921 (0.022944)
Save 5.764517 (0.150480)
Geounit Characteristics 28.376649 (0.740760)

Total 38.307478
Projected time: 38 / 500 * 285762 / 60 / 60 = 6 hours


Skip get if Geounit was created
---------------------------
Levels Queries 1.068750 (0.062079)
Shapefile attributes 0.561129 (0.032594)
Existence Query 1.292414 (0.075071)
Geometry 1.210732 (0.070326)
Save 5.702211 (0.331216)
Geounit Characteristics 7.252589 (0.421271)

Total 17.215975
Projected time (initial import): 2.7 hours
```
In other words, a 70% reduction in loading time 🎉 

### Notes

It turned out to be the creation of Geounit Characteristics that was actually taking up the majority of the time. Switching to bulk creation of the Characteristics for each Geounit shaved about 20 seconds per 500 Geounits. However, the script follows an upsert pattern which requires a preliminary query to discover whether the Characteristic already exists. Since we know that a freshly-created Geounit can't possibly have any pre-existing Characteristics, I added a flag to the `set_geounit_characteristic` function to skip the existence check if the Geounit was just created. This shaved off another 20 seconds per 500 Geounits. To make sure that we don't accidentally shoot ourselves in the foot with that flag, I made the default behavior the same as the existing function, and added a database constraint to enforce the requirement that only one Characteristic should ever exist for a Geounit/Subject pair.

We could probably gain some further speed by bulk-creating Geounits or saving up Characteristics and bulk-creating them in bigger batches, but those changes will both be more complex than this was because of the foreign key dependencies involved. 

I didn't investigate the `-n*` flags but I suspect we could likely find ways to speed those up as well.

## Testing Instructions

 * Checkout `develop`.
 * After `./scripts/console django`
 * `./manage.py setup /config/config.xml -g0 -g1 -g2` . It should take roughly an hour to reach 10% completion (Or you can make a temporary change to the [progress counter](https://github.com/PublicMapping/DistrictBuilder/blob/feature/speed-up-data-load-pt1-%23154367563/django/publicmapping/redistricting/management/commands/setup.py#L464) so that it reports progress more frequently).
 * Check out this branch, run migrations, and then rerun the command above. It should take roughly 30% as long to get to 10%.

Finishes #154367563


pcaisse (approved):
LGTM. In testing this, I saw a huge speedup. On `develop`, it ran for 68 minutes and loaded 11,203 geounits (166 geounits per minute) whereas on this branch it ran for 40 minutes and loaded 36,730 (918 geounits per minute) which represents a 5.5x speedup.

PR #497, opened 2018-01-30
Add README text to ask user to copy config.dist.xml

shreshthkhilani:
## Overview

Ask user to copy `config.dist.xml` to `config.xml` and remove the copy on provision.

### Checklist

### Notes

Should we also delete the `config.xml` file altogether? 

## Testing Instructions

 * travis


jisantuc (commented):
Yeah, let's delete it and add it to the .gitignore so no one accidentally commits secrets in the bright future where we have contributions

jisantuc (approved):

PR #498, opened 2018-01-30
Fix geounit geometry unioning during template plan creation

ddohler:
## Overview

Template plan creation using the default Virginia data wasn't working due to the way it was trying to create MultiPolygons, and it was exceeding the memory availability in the VM to boot. This fixes both problems.

This also adds database view creation to the `load_development_data` script, which is necessary for Geoserver to work.

### Checklist

### Demo

After doing these two things, the map starts to look slightly less sad.

screen shot 2018-01-30 at 1 27 28 pm

### Notes

- This makes the full `load_development_data` process take quite a bit longer because it's not erroring out on each district load. This is a good thing, but it eats into our time savings somewhat.
- I linted and removed `import *` statements from `tasks.py`; I don't think I broke anything but it's possible I did.
- There are still some errors during the Geoserver configuration step, of the form `[2018-01-30 18:20:24 +0000] [1111] [WARNING] Could not create "simple_dc_vtd" simple feature type`. It's not clear what, if anything, this is affecting.

## Testing Instructions

 * If you don't have non-simplified data already loaded, find a time when you have about 12 hours to spare and run `./scripts/load_development_data` using the non-simplified geometries.
 * If you do have non-simplified data loaded, start up a server using `./scripts/server` and then `docker compose exec vagrant_django_1 bash` inside the VM.
 * `./manage.py shell_plus` and then from Python `Plan.objects.all().delete()`
 * Load templates: `./manage.py setup config/config.xml -t` -- it should take a couple hours and not throw any errors.
 * Generate database views: `./manage.py setup config/config.xml --views` -- should be fairly quick, less than a minute.
 * Configure Geoserver: `./manage.py setup config/config.xml -G` -- should take a few minutes, may throw errors of the form above.
 * Log in as a guest, and confirm that every District Layer has an additional Plan, in addition to the "Blank" plan.
 * View one of the non-Blank plans and confirm that districts appear on the map.

Fixes #154372452


django/publicmapping/redistricting/tasks.py
```diff
@@ -483,7 +513,7 @@ def index2plan(name,
                     })
                 else:
                     logger.warn('Unable to create district %s.', district_id)
-                    logger.debug('Reason:', ex)
```
pcaisse, line 486: I feel like this kind of thing is all over the codebase. We'll tell you a problem happened but nothing else about it... :rage:

scripts/load_development_data
```diff
@@ -55,6 +55,10 @@ function load_dev_data() {
     docker-compose \
         exec -T django ./manage.py setup config/config.xml -t
 
+    echo "Creating database views"
+    docker-compose \
+        exec -T django ./manage.py setup config/config.xml --views
```
jisantuc, line 60: Sorry for the drive-by but I also wound up doing this if you can believe it 🤦‍

django/publicmapping/redistricting/tasks.py
```diff
@@ -1184,15 +1212,18 @@ def createreport(planid, stamp, request, language=None):
...
-        magnitude = map(lambda i: str(i[1]), info)
+        # magnitude = map(lambda i: str(i[1]), info)  # TODO: Reenable
 
         logger.debug('Firing web worker task.')
 
-        dispatcher = HttpDispatchTask()
+        # TODO: This was removed by Celery 4. The recommended solution is to copy-paste the Celery
+        # 3.1 implementation, but it's not clear if we still need this since it seems to be linked
+        # to BARD.
+        #dispatcher = HttpDispatchTask()
```
pcaisse: Maybe it's time to just delete this function entirely? It is indeed just for BARD. We'd have to delete the corresponding view and route as well.

pcaisse (commented):
LGTM. Going to test now.

scripts/load_development_data
ddohler: I'll rebase out before merging. 👍

pcaisse (approved):
Nice, I have districts!

PR #499, opened 2018-01-30
Fix geoserver configuration

jisantuc:
## Overview

This PR does a whole collection of different things all related to getting tiles from geoserver.

There were a few problems to solve that map pretty neatly onto the commits:

- only a subset of the subject views were created in the database when you ran setup with the `--views` flag. This was caused by an under-indented for loop.
- geoserver didn't know where to find itself. This was caused by the `PROXY_BASE_URL` not being explicitly set, so geoserver didn't know the host to ask for its copy of openlayers or other things it needed
- Database views weren't created as part of `load_development_data`. #498 also fixes this and should be merged before this PR is I think.
- The bounding box I manually set for subject layers was wrong. This caused only some of the tiles to render instead of all of the tiles, and the fix was to let geoserver figure the bbox out on its own.

There were also two ergonomics improvements:

- With batch inserting the data, django has to hang onto more data in memory, so I increased VM memory (#498 also did that it turns out)
- Because I was spending a lot of time in geoserver, I bumped the base image to a `slim` image instead of an `alpine` so `bash` and `apt` would be available. I think this is fine for development and something we should consider leaving until it's time to start caring about image size.

### Checklist

### Demo

Get

image

at

image

me

image

## Notes

If geoserver is unavailable (you'll get a `No JSON object could be decoded` from geoserver config immediately or a 503 from the service when you're using the app), `docker-compose stop`, restart the VM, and try again. If that doesn't work, talk to me. This happens sometimes, but when I was trying to create an issue, I couldn't reproduce it.

## Testing Instructions

- take down all ur stuff (`docker-compose stop`)
- bring all ur stuff back up (`docker-compose up -d`)
- Create the views: `docker-compose exec ./manage.py setup -V config/config.xml`
- Reconfigure geoserver: `docker-compose exec ./manage.py setup -G config/config.xml`
- Start drawing a plan and check out those sweet sweet tiles

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -33,14 +33,6 @@ and `config.xml`. Below is an explanation of what each of these flags does:
 
 To compile language files, use the `-l` flag.
 
-##### Static Files #####
-
-To collect static files, use the `-s` flag.
-
-##### Database Views #####
-
-The `-V` flag creates some kind of database views. It doesn't currently work though.
-
 ##### Data Setup #####
 
 The following flags are all executed as part of `scripts/load_development_data`.
@@ -73,6 +65,12 @@ accumulating record information along the way.
 The `-t` flag creates some example plans in the database to use as baselines for creating user
 plans. If it can't find information it needs to create a template, it skips that template after
 printing a warning message and doesn't fail.
+
+###### Database Views ######
+
+The `-V` flag creates the database views necessary for geoserver to show tiles. It's run as part
+of `load_development_data`.
+
 
 ###### `-G`: Configure Geoserver ######
 
--- a/README.markdown
+++ b/README.markdown
@@ -3,6 +3,8 @@ DistrictBuilder
 
 [DistrictBuilder](http://www.districtbuilder.org/) is software created by the [Public Mapping Project](http://www.publicmapping.org/resources/software).
 
+The development environment is docker-compose for services inside a Vagrant virtual machine.
+
 
 Installation
 ------------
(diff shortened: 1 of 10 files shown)


django/publicmapping/redistricting/config.py
```diff
@@ -1603,22 +1630,6 @@ def feature_template(name, title=None, alias=None, attributes=[]):
...
-                    'maxy': '%0.1f' % 5244191.63658937,
-                    'crs': 'EPSG:3785'
-                },
-                'latLonBoundingBox': {
-                    'minx': -80.61767578124999,
-                    'miny': 39.639537564366684,
-                    'maxx': -74.674072265625,
-                    'maxy': 42.5530802889558,
-                    'crs': 'EPSG:4326'
-                },
```
ddohler: I'm happy to believe these are no longer needed, but why aren't they?

django/publicmapping/redistricting/config.py
```diff
@@ -1186,6 +1189,24 @@ def configure_geoserver(self):
         @returns: A flag indicating if geoserver was configured correctly.
         """
 
+        # Set the Geoserver proxy base url
+        # This is necessary for geoserver to know where to look for its internal
+        # resources like its copy of openlayers and other things
+
+        settings = requests.get(
+            'http://geoserver.internal.districtbuilder.com:8080/geoserver/rest/settings.json'
```
ddohler: I had to add `headers=self.headers['default']` in order for this not to throw an error when running `./manage.py setup -G`

ddohler (approved):
I had to make one code change in order to get the setup script to work; other than that, this looks good, with some additional comments below that may or may not need to be addressed now:

I'm still getting warnings of the form
`[2018-01-31 18:27:29 +0000] [164] [WARNING] Could not create "va_vtd_boundaries" feature type`
when configuring Geoserver -- is that still expected?

Also the tiles seem to be misaligned compared to the real world. I don't think that necessarily needs to be fixed as part of this PR but I wanted to bring it up in case there's an easy fix.
geoservertiles

django/publicmapping/redistricting/config.py
kshepard: Yeah, I have recollections of these being required in order to get zoom levels correct (or something along those lines). This might be related to why the tiles appear shifted. One other reason for that could be the slight differences among the competing SRIDs at the time (3785 vs. 3857 vs. 900913) -- that is something that we definitely ran into when developing this, and probably have a couple workarounds in place.

django/publicmapping/redistricting/config.py
jisantuc: Interesting... when I tried this with headers, it yelled at me. That made me think that it specifically didn't want headers for GETs to the settings endpoint.

django/publicmapping/redistricting/config.py
jisantuc: Also interesting. I'll set the bbox to the max extent again and see if that fixes things. I'm rerunning the whole configuration after rebasing on develop currently anyway.

jisantuc: Er, rather, I rebased on Derek's open PR

django/publicmapping/redistricting/config.py
jisantuc: And now it's perfectly happy with headers

README.markdown
```diff
@@ -23,72 +25,53 @@ $ ./scripts/update
...
-  </Fields>
-</Shapefile>
-```
-
-that tells the management command where to look for geounits in this level, and what fields to
-care about in each shapefile record. The lowest numbered geolevel is for the smallest size, so
-if you had only counties and states, for example, counties would be `-g0`, and states would be
-`-g1`.
-
-###### `-nX`: Nesting Geolevels ######
```
pcaisse, line 66: You don't think this flag-specific documentation is useful (geolevels, templates, geoserver, database views)?

jisantuc: I thought it wasn't, since we're bundling all that up in `load_development_data`. I'll bring it back though, maybe with a reorganization that makes it more obvious that you don't _need_ to care about it, but that you can care about it if you want.

pcaisse: Okay, sounds good. My understanding was that `load_development_data` was mostly for our use to make our lives easier (VA data is hard-coded). Is that meant to be used by anyone eventually?

jisantuc: I don't know, maybe. We need to make configuration easier for people someday, and this does a lot of the processing of the configuration, so I can imagine it being used by the public with a configurable destination for where the zipped data live maybe.

jisantuc:
NEW testing instructions:

- `git pull --force` this branch
- `docker-compose exec django ./manage.py setup -G config/config.xml`
- you'll get tiles _and_ openlayers geojson _and they'll line up whoooooa_

Recent commits focused on making tiles line up and on making configuration quieter using the `quietOnNotFound` parameter

kshepard (approved):

PR #500, opened 2018-02-02
Fix most of the tests

ddohler:
## Overview

This gets us from most tests failing to most tests passing; there are about 25 left which still fail or error out.

### Checklist

## Testing Instructions

 * `./scripts/server`
 * `./scripts/test`
 * Most of them should pass! 🎉


shreshthkhilani (approved):
The changes look good to me and 25/151 tests either fail, error, or are skipped (just as claimed).

ddohler:
I decided to keep working on this since we're running low on cards, but I'm going to merge this and make another PR for the next set of fixes since some of them are a bit more involved.

PR #501, opened 2018-02-05
Fix all but two tests

ddohler:
## Overview

This gets us really close to having all tests passing; there are two remaining failures and they seem to be related.

### Checklist

### Notes

The remaining 2 failures are:
```
======================================================================
FAIL: test_get_imixed3 (redistricting.tests.test_geounit_mix.GeounitMixTestCase)
----------------------------------------------------------------------
  File "/usr/src/app/redistricting/tests/test_geounit_mix.py", line 272, in test_get_imixed3
    numunits)
AssertionError: Number of geounits outside boundary is incorrect. (62)

======================================================================
FAIL: test_get_mixed3 (redistricting.tests.test_geounit_mix.GeounitMixTestCase)
----------------------------------------------------------------------
  File "/usr/src/app/redistricting/tests/test_geounit_mix.py", line 235, in test_get_mixed3
    'Number of geounits inside boundary is incorrect. (%d)' % numunits)
AssertionError: Number of geounits inside boundary is incorrect. (62)

----------------------------------------------------------------------
```

## Testing Instructions

 * Bring up `./scripts/server`
 * `./manage.py test`

There should be a mere two failures that are the same as the ones above.

@tgilcrest there wasn't a story for this but it seemed like a useful thing to do, it took about another extra point on top of the timeboxed point we already budgeted.


jisantuc:
I don't want to merge this quite yet because we have another disappearing files problem -- when you reload your VM, geoserver loses everything that's configured, because it uses the file system for a lot of that information. Thinking about how to fix it.

jisantuc:
Oh, I commented that in the wrong PR. Sorry. Ignore that.

jisantuc (approved):

PR #502, opened 2018-02-06
Parse config.xml to .env

pcaisse:
## Overview

Move project configuration related to architecture (hosts, ports, secrets) into `.env` file. Leave all other project configuration in `config.xml`.

This also ties up a bunch of loose ends for various things that I came across while doing this work. Some examples:
* Reporting is no longer optional or configurable (always use non-BARD reporting)
* Use password for Redis (we were doing this in older iterations of DistrictBuilder, no reason why not to)
* Actually use the key-value store (Redis) configuration for calculator that uses Redis (we weren't doing that before and I don't think that calculator could have been working)
* Make languages dynamic and configurable (previously hard-coded)
* No weak passwords by default (user must provide passwords)
* Removes some settings that were just never used (`email_use_tls` and `email_subject_prefix`)
* Simplifies writing settings (eg. no longer writing `reporting_settings.py` which was used only for BARD)
* Does away with a lot of magic numbers and strings related to configuration

#### Backstory

The rationale for doing it the way it was ultimately done changed as work progressed. Initially we were going to put all the configuration in environment variables, but after thinking about it a bit more we realized that we still wanted to leverage the validation in `config.xsd`. Then the thinking was to go all in on `config.xml` and `config.xsd` so as to not have to redo validation or documentation. However, we realized that we needed to have the settings make their way into `docker-compose.yml` and the best (only?) way to do that was via environment variables. At that point it made sense to meet in the middle: use `.env` for settings that are necessary to get the containers running that wouldn't need to change much and continue to use `config.xml` for everything else. As an added bonues, `config.xml` now becomes shareable since it doesn't contain secrets.

While it does seem nice to have everything in `.env` since it would do away with a lot of complexity (`config` container, writing to settings template the result of which is imported by `settings.py`), that `.env` file would get very large, and the more settings that are configured in `.env` the more validation becomes important.

Take languages for example. We would have to have some klugey comma-separated string values like so:

```
LANGUAGES=French,English,Spanish
LANGUAGE_CODES=fr,en,sp
```

This could get unwieldy and would be easy for a user to mess up (eg. order would be important here).

Furthermore, some settings are derived from other non-project settings in `config.xml`, such as `ADJACENCY` and `COMPLEX_CLOROPLETH` so it's unclear how those could even be broken out. 

## Testing Instructions

This PR touches many files and reworks lots of things. Ideally testing would involve exercising as much of the system as possible (i.e. cloning a fresh repo and getting set up from scratch).

At a minimum, you will need to:
* `vagrant reload`
* Restart `vagrant rsync-auto` if running
* `./scripts/update`
* `./scripts/server`

Short of loading shapefiles and nesting geolevels, try running these setup commands which are quicker:
* `./manage.py setup` (tests creating a superuser)
* `./manage.py setup -t`
* `./manage.py setup --views`
* `./manage.py setup -G`

Also, log in to the app and do the following:
* Change language
* Log in as a guest and use the default template
* Log in as a registered user and make a plan
* Verify tiles are being loaded on the map
* Generate a report

Basically make sure everything still works as before.

Also, change settings and make sure they are properly validated and that they show up properly in `config_settings.py` when you run
```
docker-compose build django
```


django/publicmapping/generate_settings.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/python
+from optparse import OptionParser
```
ddohler: `optparse` is deprecated; we should be using `argparse` for new code (although I wouldn't be opposed to adding `click` as a dependency either, it's really pleasant).

django/publicmapping/generate_settings.py
```diff
@@ -0,0 +1,69 @@
...
+ERROR:
+
+    This script requires a configuration file and a schema. Please check
+    the command line arguments and try again.
+""")
+        sys.exit(1)
+
+    try:
+        config = StoredConfig(args[1], schema_file=args[0])
+    except Exception as e:
```
ddohler: It'd be great to change to `raise IOError` in `StoredConfig`, then we could narrow down the `except` here.

django/publicmapping/generate_settings.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/python
+from optparse import OptionParser
+import sys
+from redistricting import StoredConfig
```
ddohler: This and the location of `generate_settings.py` worry me a bit because they break the convention that everything inside a Django app's folder is expected to be run from within a Django context. I think this creates a potential time-bomb because it would make it easy to get into a circular-dependency pickle if we forget that this import exists and add something Django-specific to `StoredConfig`. Unfortunately the solution we have used on other projects is a bit of work because it would involve breaking out `StoredConfig` into its own Python module and either `pip`-installing it or adding it to the Python path. So I don't think anything needs to be done about this immediately, but it'd be good to refactor it soon.

ddohler (changes requested):
This looks pretty good. I ran into the same database-wipe issue you encountered, so I haven't been able to test any functionality, but I'm going to load data overnight and finish testing tomorrow morning. I made a few notes on areas that could be cleaned up.

django/publicmapping/generate_settings.py
pcaisse: Sure. I copy pasta'd most of this over from this old file: https://github.com/PublicMapping/DistrictBuilder/blob/master/django/publicmapping/setup.py

django/publicmapping/generate_settings.py
pcaisse: I don't think there's any reason this actually needs to be inside of the `redistricting` app. I just left it where it was because it was one less thing to change to update all references to it.

I think it would make more sense for `django/publicmapping/redistricting/__init__.py` to be inside of `django/publicmapping/config/`. I think it would make sense to put `config_settings.py.j2` in there as well (in fact, I have a TODO about that).

django/publicmapping/publicmapping/config_settings.py.j2
```diff
@@ -0,0 +1,50 @@
...
+        '{{ language.get('label') }}',
+    ),
+    {% endfor %}
+)
+LANGUAGE_CODE = '{{ i18n.get('default') or i18n[0].get('code') }}'
+
+BASE_MAPS = '{{ map_server.get('basemaps') }}'
+MAP_SERVER_NS = '{{ map_server.get('ns') }}'
+MAP_SERVER_NSHREF = '{{ map_server.get('nshref') }}'
+FEATURE_LIMIT = '{{ map_server.get('maxfeatures') }}'
```
ddohler: I think these quotes need to be removed; this is causing a `Cannot concatenate 'str' and 'int' objects` error when I try to run `manage.py setup -G`.

.env.sample
```diff
@@ -0,0 +1,21 @@
...
+DATABASE_PORT=5432
+KEY_VALUE_STORE_HOST=redis.internal.districtbuilder.com
+KEY_VALUE_STORE_PASSWORD=
+KEY_VALUE_STORE_PORT=6379
+KEY_VALUE_STORE_DB=0
+MAP_SERVER_HOST=geoserver.internal.districtbuilder.com
+MAP_SERVER_PORT=9091
+MAP_SERVER_USER=geoserver
+MAP_SERVER_PASSWORD=
+MAP_SERVER_ADMIN_USER=admin
```
ddohler: It doesn't seem like these are actually influencing the behavior of the Geoserver container; they change what password the Django container uses to attempt to talk to Geoserver, but the Geoserver container always seems to use Geoserver's default username/password combination, regardless of what these settings are. I think we should either put in the default here (or document it), or make Geoserver respect these settings.

django/publicmapping/generate_settings.py
jisantuc: :+1: for making `StoredConfig` its own module somewhere. I can imagine a universe where making it its own module allows us to work on config validation and wizardry in a more isolated location than in the DB main repo

django/publicmapping/publicmapping/__init__.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+REDIS_URL = 'redis://:{password}@{host}:{port}/{db}'.format(
```
jisantuc, line 3: Why does this live here instead of in settings (or the templated settings extension)?

docker-compose.yml
```diff
@@ -1,39 +1,49 @@
...
 services:
   postgres:
-    image: quay.io/azavea/postgis:2.2-postgres9.5-slim
+    env_file:
+      - .env
     environment:
-      - POSTGRES_DB=district_builder
-      - POSTGRES_USER=district_builder
-      - POSTGRES_PASSWORD=district_builder
+      - POSTGRES_DB=${DATABASE_DATABASE}
```
jisantuc, line 7: I _think_ this is probably where your data loss is happening -- the environment variable that this is getting interpolated from is [the environment in which `docker-compose` is run](https://docs.docker.com/compose/compose-file/#variable-substitution), which is the VM, and I don't think (based on my reading of the changes here) that the VM has this available, so I think you're setting these all to empty strings.

jisantuc: That might cause other issues also. There's one sort of straightforward fix, which would be sourcing `.env` as the last step of provisioning, since you tell people to copy `.env.sample` before they run `./scripts/setup` anyway

django/publicmapping/publicmapping/__init__.py
pcaisse: This is really just to avoid doing the string formatting twice (this is used in `settings.py` and `celery.py`) :man_shrugging: 

I couldn't think of a great place to put it so I just threw it in here.

docker-compose.yml
pcaisse: The variables are making their way into the `docker-compose.yml` okay. All that you need to do is have a `.env` file in the same directory: https://docs.docker.com/compose/environment-variables/#the-env-file

I verified it with `docker-compose config`:
```
vagrant@vagrant-ubuntu-trusty-64:/vagrant$ docker-compose config
services:
  celery:
    build:
      context: /vagrant/django/publicmapping
    command:
    - worker
    - --app=publicmapping
    - --uid=reporter
    - --loglevel=INFO
    entrypoint: /usr/local/bin/celery
    environment:
      ADMIN_EMAIL: admin@admin.com
      ADMIN_PASSWORD: admin
      ADMIN_USER: admin
      DATABASE_DATABASE: district_builder
      DATABASE_HOST: postgres.internal.districtbuilder.com
      DATABASE_PASSWORD: district_builder
      DATABASE_PORT: '5432'
      DATABASE_USER: district_builder
      DEBUG: "True"
      KEY_VALUE_STORE_DB: '0'
      KEY_VALUE_STORE_HOST: redis.internal.districtbuilder.com
      KEY_VALUE_STORE_PASSWORD: redis_password
      KEY_VALUE_STORE_PORT: '6379'
      MAP_SERVER_ADMIN_PASSWORD: admin
      MAP_SERVER_ADMIN_USER: admin
      MAP_SERVER_HOST: geoserver.internal.districtbuilder.com
      MAP_SERVER_PASSWORD: geoserver
      MAP_SERVER_PORT: '9091'
      MAP_SERVER_USER: geoserver
      WEB_APP_PASSWORD: django
      WEB_APP_PORT: '8080'
    links:
    - postgres:postgres.internal.districtbuilder.com
    - redis:redis.internal.districtbuilder.com
    volumes:
    - /vagrant/django/publicmapping:/usr/src/app:rw
    - reports:/opt/reports:rw
  config:
    build:
      context: /vagrant/django/publicmapping
    command:
    - generate_settings.py
    - config/config.xsd
    - config/config.xml
    entrypoint: /usr/local/bin/python
    volumes:
    - /vagrant/django/publicmapping:/usr/src/app:rw
  django:
    build:
      context: /vagrant/django/publicmapping
    command:
    - --workers=2
    - --timeout=60
    - --bind=0.0.0.0:8080
    - --reload
    - --log-level=debug
    - --access-logfile=-
    - --error-logfile=-
    - -kgevent
    - publicmapping.wsgi
    depends_on:
    - config
    entrypoint: /usr/local/bin/gunicorn
    environment:
      ADMIN_EMAIL: admin@admin.com
      ADMIN_PASSWORD: admin
      ADMIN_USER: admin
      DATABASE_DATABASE: district_builder
      DATABASE_HOST: postgres.internal.districtbuilder.com
      DATABASE_PASSWORD: district_builder
      DATABASE_PORT: '5432'
      DATABASE_USER: district_builder
      DEBUG: "True"
      KEY_VALUE_STORE_DB: '0'
      KEY_VALUE_STORE_HOST: redis.internal.districtbuilder.com
      KEY_VALUE_STORE_PASSWORD: redis_password
      KEY_VALUE_STORE_PORT: '6379'
      MAP_SERVER_ADMIN_PASSWORD: admin
      MAP_SERVER_ADMIN_USER: admin
      MAP_SERVER_HOST: geoserver.internal.districtbuilder.com
      MAP_SERVER_PASSWORD: geoserver
      MAP_SERVER_PORT: '9091'
      MAP_SERVER_USER: geoserver
      WEB_APP_PASSWORD: django
      WEB_APP_PORT: '8080'
    expose:
    - '8080'
    links:
    - postgres:postgres.internal.districtbuilder.com
    - redis:redis.internal.districtbuilder.com
    - geoserver:geoserver.internal.districtbuilder.com
    volumes:
    - /vagrant/django/publicmapping:/usr/src/app:rw
    - /vagrant/sld:/opt/sld:rw
    - data:/data:rw
    - reports:/opt/reports:rw
  geoserver:
    build:
      context: /vagrant/geoserver
    environment:
      ADMIN_EMAIL: admin@admin.com
      ADMIN_PASSWORD: admin
      ADMIN_USER: admin
      DATABASE_DATABASE: district_builder
      DATABASE_HOST: postgres.internal.districtbuilder.com
      DATABASE_PASSWORD: district_builder
      DATABASE_PORT: '5432'
      DATABASE_USER: district_builder
      DEBUG: "True"
      KEY_VALUE_STORE_DB: '0'
      KEY_VALUE_STORE_HOST: redis.internal.districtbuilder.com
      KEY_VALUE_STORE_PASSWORD: redis_password
      KEY_VALUE_STORE_PORT: '6379'
      MAP_SERVER_ADMIN_PASSWORD: admin
      MAP_SERVER_ADMIN_USER: admin
      MAP_SERVER_HOST: geoserver.internal.districtbuilder.com
      MAP_SERVER_PASSWORD: geoserver
      MAP_SERVER_PORT: '9091'
      MAP_SERVER_USER: geoserver
      WEB_APP_PASSWORD: django
      WEB_APP_PORT: '8080'
    links:
    - postgres:postgres.internal.districtbuilder.com
    ports:
    - published: 9091
      target: 8080
    volumes:
    - /vagrant/geoserver/data_dir:/data:rw
    - /vagrant/sld:/data/sld:rw
  nginx:
    command: /bin/bash -c "envsubst < /etc/nginx/conf.d/default.template > /etc/nginx/conf.d/default.conf
      && nginx -g 'daemon off;'"
    depends_on:
    - django
    environment:
      ADMIN_EMAIL: admin@admin.com
      ADMIN_PASSWORD: admin
      ADMIN_USER: admin
      DATABASE_DATABASE: district_builder
      DATABASE_HOST: postgres.internal.districtbuilder.com
      DATABASE_PASSWORD: district_builder
      DATABASE_PORT: '5432'
      DATABASE_USER: district_builder
      DEBUG: "True"
      KEY_VALUE_STORE_DB: '0'
      KEY_VALUE_STORE_HOST: redis.internal.districtbuilder.com
      KEY_VALUE_STORE_PASSWORD: redis_password
      KEY_VALUE_STORE_PORT: '6379'
      MAP_SERVER_ADMIN_PASSWORD: admin
      MAP_SERVER_ADMIN_USER: admin
      MAP_SERVER_HOST: geoserver.internal.districtbuilder.com
      MAP_SERVER_PASSWORD: geoserver
      MAP_SERVER_PORT: '9091'
      MAP_SERVER_USER: geoserver
      WEB_APP_PASSWORD: django
      WEB_APP_PORT: '8080'
    image: nginx:1.12
    ports:
    - published: 8080
      target: 8080
    volumes:
    - /vagrant/nginx:/etc/nginx/conf.d:rw
    - /vagrant/django/publicmapping/static:/opt/static:rw
    - /vagrant/sld:/opt/sld:rw
    - reports:/opt/reports:rw
  postgres:
    environment:
      ADMIN_EMAIL: admin@admin.com
      ADMIN_PASSWORD: admin
      ADMIN_USER: admin
      DATABASE_DATABASE: district_builder
      DATABASE_HOST: postgres.internal.districtbuilder.com
      DATABASE_PASSWORD: district_builder
      DATABASE_PORT: '5432'
      DATABASE_USER: district_builder
      DEBUG: "True"
      KEY_VALUE_STORE_DB: '0'
      KEY_VALUE_STORE_HOST: redis.internal.districtbuilder.com
      KEY_VALUE_STORE_PASSWORD: redis_password
      KEY_VALUE_STORE_PORT: '6379'
      MAP_SERVER_ADMIN_PASSWORD: admin
      MAP_SERVER_ADMIN_USER: admin
      MAP_SERVER_HOST: geoserver.internal.districtbuilder.com
      MAP_SERVER_PASSWORD: geoserver
      MAP_SERVER_PORT: '9091'
      MAP_SERVER_USER: geoserver
      POSTGRES_DB: district_builder
      POSTGRES_PASSWORD: district_builder
      POSTGRES_USER: district_builder
      WEB_APP_PASSWORD: django
      WEB_APP_PORT: '8080'
    healthcheck:
      interval: 15s
      retries: 3
      test:
      - CMD-SHELL
      - PGPASSWORD=district_builder
      - pg_isready
      - -U
      - district_builder
      timeout: 5s
    image: quay.io/azavea/postgis:2.2-postgres9.5-slim
  redis:
    command: redis-server --requirepass redis_password
    environment:
      ADMIN_EMAIL: admin@admin.com
      ADMIN_PASSWORD: admin
      ADMIN_USER: admin
      DATABASE_DATABASE: district_builder
      DATABASE_HOST: postgres.internal.districtbuilder.com
      DATABASE_PASSWORD: district_builder
      DATABASE_PORT: '5432'
      DATABASE_USER: district_builder
      DEBUG: "True"
      KEY_VALUE_STORE_DB: '0'
      KEY_VALUE_STORE_HOST: redis.internal.districtbuilder.com
      KEY_VALUE_STORE_PASSWORD: redis_password
      KEY_VALUE_STORE_PORT: '6379'
      MAP_SERVER_ADMIN_PASSWORD: admin
      MAP_SERVER_ADMIN_USER: admin
      MAP_SERVER_HOST: geoserver.internal.districtbuilder.com
      MAP_SERVER_PASSWORD: geoserver
      MAP_SERVER_PORT: '9091'
      MAP_SERVER_USER: geoserver
      WEB_APP_PASSWORD: django
      WEB_APP_PORT: '8080'
    image: redis:3.2-alpine
    ports:
    - published: 6379
      target: 6379
version: '3.3'
volumes:
  data: {}
  reports: {}
```
Plus if everything was blank nothing would work because hosts and ports are defined there.

ddohler (changes requested):
I ran into a couple more issues after loading development data. My geoserver is also no longer returning tiles, but I don't know whether that's related or not.

docker-compose.yml
jisantuc: Interesting. Nevermind then!

django/publicmapping/publicmapping/__init__.py
pcaisse: I broke it out into a separate module (not in `settings.py`) because I running into an issue with `settings.py` being a celery dependency and the order in which settings were being built out. Something like celery being brought up before django and then django not having the extended settings yet. It also doesn't _need_ to be in settings. It's not really a setting, just a formatted string.

.env.sample
pcaisse: > I think we should either put in the default here (or document it), or make Geoserver respect these settings.

I would be good with either of those, though the latter seems preferable if it's not too difficult.

django/publicmapping/publicmapping/__init__.py
ddohler: Oh, yeah, docker-compose doesn't wait for dependencies to be "ready" before bringing them up, so what's happening is that `django` and `celery` are launched immediately after `config` is, before `config` has had a chance to write out the config file. So it might be easier just to duplicate this, if it's the only thing that `celery.py` needs, then `celery` might not have to depend on `django` at all?

docker-compose.yml
pcaisse: But, relatedly, I didn't push this commit to properly map the environment variables until recently (after @ddohler started testing): https://github.com/PublicMapping/DistrictBuilder/pull/502/commits/9944ba8518cf5209ca35beb55248d0f0bb103bad

My theory is that before pushing that commit, we were connecting to the wrong database, whatever the default was.

django/publicmapping/publicmapping/__init__.py
pcaisse: To be honest, I'm not entirely sure I understand what the dependencies between Celery and Django are.

In `celery.py` we refer to Django settings as per the docs' recommendation. Django uses the defined Celery app to queue up tasks. The containers themselves don't actually depend on each other at all. I think they just share code but as long as a Celery worker is running by the time a task is called, things will work. Maybe `django` should depend on `celery`?

django/publicmapping/generate_settings.py
pcaisse: I'm going to hold off on doing anything about this for now.

.env.sample
jisantuc: Can confirm geoserver is ignoring these :(

django/publicmapping/generate_settings.py
kshepard: I also agree that splitting this out to its own module would be a nice improvement. Want to add a story for it, so the idea doesn't get lost?

jisantuc:
Another issue (I haven't found where in the code this is, _but_) is that routing to geoserver is broken on this branch. For the browser to find geoserver, requests need to go through `:8080/geoserver/...` so nginx can find it. Right now it's sending requests to `geoserver.internal.districtbuilder.com`, which is the value of `MAP_SERVER_HOST`, and the browser has no idea what that is.

django/publicmapping/generate_settings.py
pcaisse: I ended up doing most all of the work for this to get languages working after rebasing (see 4893d20a0b62241339347f9916f4332e55ccf072). I like it a lot better as a package that gets installed. The only remaining work for it would be to actually put the package in its own repo.

pcaisse:
@ddohler @jisantuc The issues with missing data and geoserver tiles not loading should be solved now.

pcaisse:
Last night I ran `./scripts/configure` in its entirety and it succeeded. Everything seems to be working :tada:

jisantuc (approved):
Can confirm that I have tiles and a functioning app on this branch. The only thing left that I can see that could matter is that geoserver still doesn't seem to take the user/password from the `.env`, but I think it's fine to open a story for that and close this out. Interested to see if @ddohler agrees

.travis.yml
```diff
@@ -2,6 +2,13 @@ services:
   - docker
 
 install:
+  - cp .env.sample .env
+  - sed -i 's/WEB_APP_PASSWORD=/WEB_APP_PASSWORD=peter/' .env
```
ddohler: This should maybe be `travis`?

ddohler (approved):
I had one last comment but this looks good otherwise, and I agree that we can hold off on fixing the Geoserver password issue for the moment.

:+1: After squashing

PR #507, opened 2018-02-08
Break out config file loading into separate script

ddohler:
## Overview

This should make deployment easier by allowing data loading to happen
without dropping the database and running migrations.

### Checklist

- [ ] ~Files changed in the PR have been `yapf`-ed for style violations~

### Notes

I think this is what we discussed but I'm assigning extra reviewers to make sure this works for everyone.

## Testing Instructions

 * (Probably) comment out `clear_database` from `scripts/configure` if you already have data
 * `vagrant ssh` and run both `scripts/configure` and `scripts/load_configured_data`, and make sure they both work as expected.

@tnation14 isn't on the repo yet but I invited him and will add him as a reviewer when he confirms.

### DIFF
--- a/scripts/configure_va_demo
+++ b/scripts/configure_va_demo
@@ -1,0 +1,58 @@
+#!/bin/bash
+set -e
+
+if [[ -n "${DB_DEBUG}" ]]; then
+    set -x
+fi
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+function usage() {
+
+    echo -n \
+"Usage: $(basename "$0") [config.xml]
+
+Configure the application with demo data
+"
+}
+
+function fetch_dev_data() {
+    echo "Creating directory to download data into in django container"
+    docker-compose \
+        exec -T django mkdir -p /data
+
+    echo "Downloading shapefile"
+    docker-compose \
+        exec -T django \
+        wget -q \
+            -O /data/districtbuilder_data.zip \
+            https://s3.amazonaws.com/districtbuilderdata/VA_data.zip
+}
+
+function recreate_database() {
+    echo "Drop district_builder database"
+    docker-compose \
+        exec -T postgres gosu postgres dropdb --if-exists district_builder
+
+    echo "Create district_builder database"
+    docker-compose \
+        exec -T postgres gosu postgres createdb district_builder
+
+    echo "Running migrations"
+    docker-compose \
+        exec -T django ./manage.py migrate
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]
+then
+    if [ "${1:-}" = "--help" ]
+    then
+        usage
+    else
+        configPath="config/${1:-config.xml}"
+        docker-compose up -d postgres django
+        #fetch_dev_data
+        #recreate_database
+        "${DIR}/load_configured_data" "${configPath}"
+    fi
+    exit
+fi
--- a/README.markdown
+++ b/README.markdown
@@ -28,8 +28,19 @@ $ cp django/publicmapping/.env.sample django/publicmapping/.env
 $ ./scripts/setup
 $ vagrant ssh
 $ ./scripts/update
-$ ./scripts/configure
 ```
+
+If you want to get DistrictBuilder up and running quickly with demo data, you can then run
+```bash
+$ ./scripts/configure_va_demo
+```
+
+Otherwise, you'll need to provide your own shapefiles and config.xml file, and then run
+```bash
+$ ./scripts/load_configured_data
+```
+
+More detailed instructions on loading your own data can be found below.
 
 :zzz:
 
@@ -57,9 +68,9 @@ Ease of interacting with the configuration file is a planned area for future dev
 with docker installed. `vagrant ssh` gets you into the virtual machine so you can run commands.
 From there, running `./scripts/update` builds containers. The rest of the setup happens either
 directly or indirectly through a setup management command. To get started, run
-`./script/setup`, followed by `vagrant ssh`, followed by `./scripts/update`.
+`./scripts/setup`, followed by `vagrant ssh`, followed by `./scripts/update`.
  
-Then, run `./scripts/configure`. It is not fast. Currently, it takes several hours, with the exact
+Then, run `./scripts/configure_va_demo`. It is not fast. Currently, it takes several hours, with the exact
 time depending on hardware. We are working on ways to improve the speed of loading data.
 
 The script will do several things
--- a/deployment/scripts/load_configured_data.sh
+++ b/deployment/scripts/load_configured_data.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+set -e
+
+if [[ -n "${DB_DEBUG}" ]]; then
+    set -x
+fi
+
+function usage() {
+
+    echo -n \
+"Usage: $(basename "$0") [config file path]
+
+Configure the application with demo data loaded from /data/districtbuilder_data.zip
+"
+}
+
+function load_configured_data() {
+    echo "Unzipping shapefile"
+    docker-compose \
+        exec -T django \
+        unzip -o /data/districtbuilder_data.zip -d /data
+
+    echo "Loading shapefiles into database"
+    docker-compose \
+        exec -T django ./manage.py setup "${configPath}" -g0 -g1 -g2
+
+    echo "Nesting geolevels"
+    docker-compose \
+        exec -T django ./manage.py setup "${configPath}" -n0 -n1 -n2
+
+    echo "Creating template plans"
+    docker-compose \
+        exec -T django ./manage.py setup "${configPath}" -t
+
+    echo "Creating database views"
+    docker-compose \
+        exec -T django ./manage.py setup "${configPath}" --views
+
+    echo "Configuring geoserver"
+    docker-compose \
+        exec -T django ./manage.py setup "${configPath}" -G
+
+    echo "Creating translation files"
+    docker-compose \
+        exec -T django ./manage.py setup "${configPath}" -l
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]
+then
+    if [ "${1:-}" = "--help" ]
+    then
+        usage
+    else
+        configPath=${1:-config/config.xml}
+        docker-compose up -d postgres django
+        load_configured_data
+    fi
+    exit
+fi
+
--- a/django/publicmapping/publicmapping/settings.py
+++ b/django/publicmapping/publicmapping/settings.py
@@ -15,6 +15,7 @@ NOTE: This settings file should not be changed!
 
 import os
 import logging.config
+import logging
 from . import REDIS_URL
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
@@ -232,4 +233,9 @@ REPORTS_ENABLED = 'CALC'
 # NOTE: Leave this at the end of the file!
 # These settings are generated based on config.xml
 # and allow for modifiying/overriding the default settings
-from publicmapping.config_settings import *
+try:
+    from publicmapping.config_settings import *
+except ImportError:
+    logger = logging.getLogger(__name__)
+    logger.error('Could not find config_settings; double-check the README for missed setup steps')
+    raise
--- a/scripts/load_configured_data
+++ b/scripts/load_configured_data
@@ -8,40 +8,17 @@ fi
 function usage() {
 
     echo -n \
-"Usage: $(basename "$0") [config.xml]
+"Usage: $(basename "$0") [config file path]
 
-Configure the application with demo data
+Configure the application with demo data loaded from /data/districtbuilder_data.zip
 "
 }
 
-function load_dev_data() {
-    echo "Creating directory to download data into in django container"
-    docker-compose \
-        exec -T django mkdir -p /data
-
-    echo "Downloading shapefile"
-    docker-compose \
-        exec -T django \
-        wget -q \
-            -O /data/VA_data.zip \
-            https://s3.amazonaws.com/districtbuilderdata/VA_data.zip
-
+function load_configured_data() {
     echo "Unzipping shapefile"
     docker-compose \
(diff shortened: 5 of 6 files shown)


kshepard (changes requested):
This looks pretty close to what I was expecting. `load_configured_data` looks great. I wonder if `configure` should be split out a little more though. Just thinking from the perspective of someone who wants to deploy a non-VA instance. I think the deployment script will want to run the `clear_database` function, though not the rest of `configure`. And on the user's development environment, they will want to do something similar -- they will be manually assembling/downloading the data zip file, and would want to run the `clear_database` function, followed by `load_configured_data`. So maybe the `clear_database` function just needs to be split out into it's own script?

scripts/configure
```diff
@@ -23,14 +24,11 @@ function load_dev_data() {
...
+            -O /data/districtbuilder_data.zip \
             https://s3.amazonaws.com/districtbuilderdata/VA_data.zip
+}
 
-    echo "Unzipping shapefile"
-    docker-compose \
-        exec -T django \
-        unzip -o /data/VA_data.zip -d /data
-
+function clear_database() {
```
pcaisse: I think I like `recreate_database` better to make it clear the database will be dropped /nitpick

pcaisse (commented):
I see there as being 4 steps you might want or need to do separately:

1. Recreate (drop and create) the database
1. Run migrations
1. Fetch and load data
1. Configure

I don't really like the "scripts to rule them all" pattern because each step is a bash script which adds a lot of boilerplate (eg. `set -e` at the top of each and every file) and they're not easily composable. The idea of a this nice mapping of "things you always want to do on a project" (eg. boostrap, setup, update) and a bash script breaks down pretty quickly as we're seeing. It would be nice to have granular control over what steps happen and how to group them beyond writing a specific script that groups all of those steps or running multiple scripts.

scripts/configure
```diff
@@ -23,14 +24,11 @@ function load_dev_data() {
     docker-compose \
         exec -T django \
         wget -q \
-            -O /data/VA_data.zip \
+            -O /data/districtbuilder_data.zip \
             https://s3.amazonaws.com/districtbuilderdata/VA_data.zip
```
tnation14: I think we might want to parametrize the `DISTRICTBUILDER_DATA_URL` as well. Additionally, did we agree to assume that the data would be stored on S3? If so, I'd rather use `aws s3 cp` over `wget` so that the data bucket doesn't have to be publicly accessible.

tnation14 (commented):
Looks good!

> The idea of a this nice mapping of "things you always want to do on a project" (eg. boostrap, setup, update) and a bash script breaks down pretty quickly as we're seeing

I think that although STRTA vary in implementation across projects, the high level idea of needing to perform certain tasks to setup your dev environment has held true for the most part. This particular project is causing some complexity because we're trying to use the same scripts to setup Dev and Prod environments, which is generally not the case. That said, it may be worth it to move this functionality into separate deployment scripts, and just let the dev STRTA remain as they previously were.

scripts/configure
ddohler: My recollection is that we ended up agreeing that the data would be placed in a designated local directory, which is why I opted not to parametrize this, because I think this script is unique to development setup. I think the same thing would apply to the `wget` vs. `s3` choice, but if I'm misremembering, let me know.

ddohler:
I think @tnation14 's thought about avoiding mixing STRTA and deployment is a good one, so I'm going to refactor to add these as scripts under `/deployment` as well as `/scripts`. I'll leave things split out on the `scripts` side as well to provide ourselves with a starting point for developing and deploying multiple instances. That will result in some duplication between the deployment version of `load_configured_data` and the STRTA version, but I think that's fine because they'll probably evolve away from each other as we develop our deployment processes and this will give them the flexibility to grow independently.

scripts/configure
tnation14: I don't think you're misremembering, that sounds right to me.

ddohler:
Okay, I've updated this. Looking at the development scripts issue, I noticed that we're not really following what STRTA says that our scripts should be doing, which I think led to some of the database work being done in the "wrong" place. I've refactored with the following changes:
- `load_configured_data` has been copied to `deployment/scripts/load_configured_data.sh`
- `scripts/setup` has been renamed to `scripts/bootstrap`
- `scripts/update` has been renamed to `scripts/setup` and now includes running migrations after containers are built.
- `scripts/configure` has been renamed `scripts/configure_va_demo`
- README and `.travis.yml` have been updated appropriately

This should address all the use cases we're considering, I think:
- In deployment, we can place the shapefile in a directory and run `deployment/scripts/load_configured_data.sh` on it. If we need to make changes to this script for deployment, we can, because it's a separate script.
- In local development, setting up the demo VA data is basically the same as before:
    1. `scripts/bootstrap`
    2. `scripts/setup`
    3. `scripts/configure_va_demo`
- But if you want to do development with a different state and config file, then you can:
    1. `scripts/bootstrap`
    2. `scripts/setup` <-- Leaves you with a migrated and ready database
    3. <Manually set up your config.xml and shapefile>
    4. `scripts/load_configured_data`

One thing that I thought about doing, but didn't, is having `scripts/setup` wipe the database before running migrations, and adding a separate `scripts/update` that just runs migrations. I think this could become useful if we find ourselves frequently switching between development on different DistrictBuilder deployments. However, we're not quite there yet, and adding functionality that could create 12-hour re-configuration headaches for anyone who accidentally runs it didn't seem like a step forward. If we decide we really need that functionality then we can add a warning to the script, but for the moment it didn't seem like it would be very useful.

This is ready for another look, I think.

.travis.yml
```diff
@@ -9,7 +9,7 @@ install:
   - sed -i 's/KEY_VALUE_STORE_PASSWORD=/KEY_VALUE_STORE_PASSWORD=redis/' .env
   - sed -i 's/MAP_SERVER_ADMIN_PASSWORD=/MAP_SERVER_ADMIN_PASSWORD=geoserver/' .env
   - cp django/publicmapping/config/config.dist.xml django/publicmapping/config/config.xml
-  - ./scripts/update
+  - ./scripts/setup
```
tnation14: Minor change, but according to https://github.com/azavea/architecture/blob/master/doc/arch/adr-0000-scripts-to-rule-them-all.md, I think we want the script that provisions the VM to be `scripts/setup`.The script that builds containers/runs migrations can stay named `scripts/update`.

tnation14:
Not sure if this within the scope of the PR or just a function of setting up a new dev environment, but I can't seem to get the `django` container to start:

```bash
$ vagrant ssh
vagrant@vagrant-trusty-64:/vagrant $ ./scripts/setup
...
Running migrations
  File "./manage.py", line 21, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 308, in execute
    settings.INSTALLED_APPS
  File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 56, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 41, in _setup
    self._wrapped = Settings(settings_module)
  File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 110, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
  File "/usr/local/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/usr/src/app/publicmapping/settings.py", line 235, in <module>
    from publicmapping.config_settings import *
ImportError: No module named config_settings
```

Even after rebuilding the container, `config_settings.py` isn't there, but the container throws a different error:

```bash
django_1     | Traceback (most recent call last):
django_1     |   File "/usr/local/lib/python2.7/site-packages/gunicorn/arbiter.py", line 578, in spawn_worker
django_1     |     worker.init_process()
django_1     |   File "/usr/local/lib/python2.7/site-packages/gunicorn/workers/ggevent.py", line 190, in init_process
django_1     |     super(GeventWorker, self).init_process()
django_1     |   File "/usr/local/lib/python2.7/site-packages/gunicorn/workers/base.py", line 126, in init_process
django_1     |     self.load_wsgi()
django_1     |   File "/usr/local/lib/python2.7/site-packages/gunicorn/workers/base.py", line 135, in load_wsgi
django_1     |     self.wsgi = self.app.wsgi()
django_1     |   File "/usr/local/lib/python2.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
django_1     |     self.callable = self.load()
django_1     |   File "/usr/local/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 65, in load
django_1     |     return self.load_wsgiapp()
django_1     |   File "/usr/local/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 52, in load_wsgiapp
django_1     |     return util.import_app(self.app_uri)
django_1     |   File "/usr/local/lib/python2.7/site-packages/gunicorn/util.py", line 352, in import_app
django_1     |     __import__(module)
django_1     |   File "/usr/local/lib/python2.7/site-packages/gevent/builtins.py", line 93, in __import__
django_1     |     result = _import(*args, **kwargs)
django_1     |   File "/usr/src/app/publicmapping/wsgi.py", line 16, in <module>
django_1     |     application = get_wsgi_application()
django_1     |   File "/usr/local/lib/python2.7/site-packages/django/core/wsgi.py", line 13, in get_wsgi_application
django_1     |     django.setup(set_prefix=False)
django_1     |   File "/usr/local/lib/python2.7/site-packages/django/__init__.py", line 22, in setup
django_1     |     configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
django_1     |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 56, in __getattr__
django_1     |     self._setup(name)
django_1     |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 41, in _setup
django_1     |     self._wrapped = Settings(settings_module)
django_1     |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 129, in __init__
django_1     |     raise ImproperlyConfigured("The SECRET_KEY setting must not be empty.")
django_1     | ImproperlyConfigured: The SECRET_KEY setting must not be empty.
django_1     | [2018-02-13 21:02:17 +0000] [8] [INFO] Worker exiting (pid: 8)
django_1     | [2018-02-13 21:02:17 +0000] [9] [INFO] Worker exiting (pid: 9)
django_1     | [2018-02-13 21:02:17 +0000] [1] [INFO] Shutting down: Master
django_1     | [2018-02-13 21:02:17 +0000] [1] [INFO] Reason: Worker failed to boot.
```

.travis.yml
kshepard: For your errors above, I wonder if you need to do the couple steps mentioned in the README:
```
$ cp django/publicmapping/config/config.dist.xml django/publicmapping/config/config.xml
$ # Copy .env file and add password/modify as necessary
$ # NOTE: Leave MAP_SERVER_ADMIN_PASSWORD the same for now
$ cp django/publicmapping/.env.sample django/publicmapping/.env
```

pcaisse: I wonder if it might be helpful to have a `Makefile` that we could use where the targets would have dependencies on these configuration files that need to be in place to run the app.

Alternatively, we could make the `settings.py` file a bit smarter by catching exceptions and outputting more helpful messaging when `docker-compose up` fails (eg. "Did you forget to do a build?").

ddohler: @tnation14 good call; I had forgotten we had our own STRTA ADR and was going off the GitHub spec, which includes a `bootstrap` script, whereas ours doesn't. I've shifted everything back to use `setup` and `update`.

ddohler: @pcaisse I've gone ahead and added a try/except to settings which prints a more helpful message if `config_settings.py` isn't found.

kshepard (approved):

.travis.yml
tnation14: Turns out that `config_settings.py` was getting overridden by `docker-compose` volume mounts, so Django couldn't find the file during initialization. I was able to get the script to succeed from a clean environment by adding this command to `scripts/update`:

```diff
diff --git a/scripts/update b/scripts/update
index 6301c3bc..bda76076 100755
--- a/scripts/update
+++ b/scripts/update
@@ -31,6 +31,14 @@ then
         usage
     else
         build_containers
+
+        # Write config_settings to the filesystem so it doesn't get
+        # overwritten by docker-compose volume mounts
+        docker-compose run --rm --entrypoint "bash -c" django \
+            "python -m district_builder_config.generate_settings \
+                config/config.xsd \
+                config/config.xml \
+                publicmapping/config_settings.py"
         docker-compose up -d postgres django
         run_migrations
     fi
```

kshepard: Cool, it seems this is the same issue discussed here: https://github.com/PublicMapping/DistrictBuilder/pull/512#pullrequestreview-96657452

That's great that you likely came up with a fix.

scripts/configure_va_demo
```diff
@@ -0,0 +1,58 @@
...
+        exec -T django \
+        wget -q \
+            -O /data/districtbuilder_data.zip \
+            https://s3.amazonaws.com/districtbuilderdata/VA_data.zip
+}
+
+function recreate_database() {
+    echo "Drop district_builder database"
+    docker-compose \
+        exec -T postgres gosu postgres dropdb district_builder
```
tnation14: Might be good to add an `--if-exists` flag to `dropdb` here.

tnation14 (approved):
All good pending the two changes I suggested

.travis.yml
ddohler: I'm going to add this into this PR since the issue seems to be afflicting Travis.

ddohler:
I'm going to go ahead and merge so that I can rebase #513 on this. The changes to `scripts/update` allowed Travis to make it through to running tests, so I think there's a reasonable chance that that PR will be able to pass after these changes are incorporated.

PR #508, opened 2018-02-09
Fix 500s on moving geounits between districts

jisantuc:
## Overview

This PR makes it so that moving geounits between plans doesn't throw a 500. Because
the reason it threw a 500 also would have caused other 500s, I changed other places
where we tried to access the `REQUEST` property of a `WSGIRequest` as well.

### Checklist

## Testing Instructions

 * claim you forgot your password and enter your username
 * observe that you see your password hint
 * log in for real
 * pick a template plan and start moving geounits around
 * observe that it is in fact possible to do the previous testing step


pcaisse (approved):
Ohhhh... so _that's_ what the app does!

PR #509, opened 2018-02-12
Fix regression that broke reporting

pcaisse:
## Overview

The shared volume must be used so that Django, Celery, and nginx can share static files.

## Testing Instructions

 * `./scripts/server`
 * Try to generate a report and notice it now succeeds

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,7 +46,7 @@ services:
       - "${WEB_APP_PORT}"
     volumes:
       - ./django/publicmapping/:/usr/src/app/
-      - ./reports:/opt/reports
+      - reports:/opt/reports
     entrypoint: /usr/local/bin/gunicorn
     command:
       - "--workers=2"


jisantuc:
Can you explain the issue a little more? Fixing exports, I noticed I had to add `reports` to `rsync_exclude`, or sometimes `rsync-auto` would clobber my file before I could consume it. Is this similar, or something else?

pcaisse:
@jisantuc I think this was just an accident introduced in this PR: https://github.com/PublicMapping/DistrictBuilder/pull/499/files#diff-4e5e90c6228fd48698d074241c2ba760

The dot slash before 'reports' (`- ./reports:/opt/reports`) makes it a directory relative to where `docker-compose.yml` is. Just referring to `- reports:/opt/reports` makes it use the shared `reports` volume defined at the bottom of the file, which is necessary so that the different containers can see the resultant HTML files that are written by the report generation task.

jisantuc:
Ah I see interesting. Lemme see if my exports PR works with the `reports` named volume, a minute ago it didn't seem to be but I've fixed other things since then. It's possible this needs to be two different things at once for different reasons :tada:

jisantuc (approved):
Confirmed this is correct / that django and celery understand how to talk to each other via volumes

PR #510, opened 2018-02-12
Make exports work

jisantuc:
## Overview

This PR fixes exports.

### Checklist

### Demo

Check out this shapefile I exported: [asdfasdf3.zip](https://github.com/PublicMapping/DistrictBuilder/files/1717947/asdfasdf3.zip)

And this district index file! [asdfasdf3(1).zip](https://github.com/PublicMapping/DistrictBuilder/files/1717950/asdfasdf3.1.zip)

### Notes

Metadata is being written "correctly" from a certain perspective, but QGIS doesn't seem to like the spatial reference definition. I think fixing that is a out-of-scope here, so moving along. For confirmation that the `.xml` file in the `shapefile bundle is "correct", you can compare it to the `.shp.xml` file in some TIGER shapefiles, [like Congressional districts](https://www.census.gov/geo/maps-data/data/cbf/cbf_cds.html), if you want.

Also roundtrips don't work because upload is messed up :(. I created a story in PT.

## Testing Instructions

 * rebuild django and celery (there are new dependencies on `dict2xml`, `fiona`, and security-patched `django`)
 * Create a new plan from a template (or from blank and add at least one district)
 * Try to export a district index file
 * It should succeed and you should get a csv
 * Try to export a shapefile
 * Save the shapefile
 * Open it in qgis (you'll have to tell it the spatial reference -- it will correctly guess WGS84 but you'll have to tell it it's right) and confirm it looks like Virginia


nginx/default.template
```diff
@@ -12,6 +12,8 @@ server {
 
   location / {
     proxy_pass http://web;
+    client_max_body_size 20M
+    ;
```
pcaisse: Any reason why there's a line break before this semicolon?

pcaisse (approved):
I was able to download the CSV and the shapefile and view it in QGIS (had to `vagrant reload` first though, so I can confirm the rsync-ing was clobbering those files!)

nginx/default.template
jisantuc: Yes -- clumsiness

PR #511, opened 2018-02-12
Fix reporting hack

pcaisse:
## Overview

This rolls back a kludgey fix to get reporting working and instead changes the default configuration to use the correct fully-qualified module name.

### Notes

When I initially made this change I didn't understand that for better or for worse `config.xml` points to files on the file system (I didn't really understand how the module name was making its way into the database). It seems better to change the configuration instead of trying to "fix" the incorrect config downstream. 

## Testing Instructions

 * `cp django/publicmapping/config/config.dist.xml django/publicmapping/config/config.xml` 
 * `./scripts/server`
 * `docker-compose exec -T django ./manage.py setup -f config/config.xml`
 * Generate a report and observe that it works

### DIFF
--- a/django/publicmapping/config/config.dist.xml
+++ b/django/publicmapping/config/config.dist.xml
@@ -132,7 +132,7 @@
         <ScoreFunctions>
             <!-- A district score that returns a percentage -->
             <ScoreFunction id="district_blkvap_percent" type="district"
-                calculator="publicmapping.redistricting.calculators.Percent"
+                calculator="redistricting.calculators.Percent"
                 label="Black VAP">
                 <SubjectArgument name="numerator" ref="vapblk" />
                 <SubjectArgument name="denominator" ref="vap" />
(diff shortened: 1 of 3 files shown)


django/publicmapping/redistricting/tasks.py
```diff
@@ -1195,16 +1195,11 @@ def createcalculatorreport(planid, stamp, request, language=None):
...
                                       plan.version, stamp)
-        htmlfile = open(
-            '%s/%s.html' % (
-                tempdir,
-                filename,
-            ), mode='w', encoding='utf=8')
-        htmlfile.write(html)
-        htmlfile.close()
+        file_path = '%s/%s.html' % (settings.REPORTS_ROOT, filename)
+        with open(file_path, mode='w', encoding='utf=8') as htmlfile:
```
ddohler: Typo, `utf=8`

django/publicmapping/config/config.dist.xml
```diff
@@ -132,7 +132,7 @@
         <ScoreFunctions>
             <!-- A district score that returns a percentage -->
             <ScoreFunction id="district_blkvap_percent" type="district"
-                calculator="publicmapping.redistricting.calculators.Percent"
+                calculator="redistricting.calculators.Percent"
```
ddohler, line 135: This is the opposite of the change I had to make to get tests to pass; I wonder why tests need `publicmapping` prepended but reports do not.

ddohler (approved):
My most recent data load apparently didn't completely work, so I haven't been able to test, but this looks good except for one minor typo.

PR #512, opened 2018-02-13
Install district_builder_config Python package from separate repo

pcaisse:
## Overview

Installs `district_builder_config` package from newly created GitHub repo: https://github.com/PublicMapping/district-builder-config

This keeps the configuration code separate and demonstrates there are no dependencies on Django. 

### Checklist

## Testing Instructions

This PR doesn't include any functional changes.

 * `docker-compose build`. The `district_builder_config` package should download and install successfully and the `generate_settings` script should also run successfully.
 * `docker-compose up` should work (app should run normally)
 * Run `docker-compose exec -T django ./manage.py setup -f config/config.xml` just for fun and see that it works (this demonstrates the setup command importing `StoredConfig` from the installed package correctly).


ddohler (approved):
This looks good and works as far as I can tell. However, I ran into an issue where `config_settings.py` disappeared and absolutely refused to come back, both on this branch and `develop`, despite multiple invocations of `docker-compose build` and `docker-compose rm django`. I ultimately figured out that whatever is at `/vagrant/django/publicmapping/publicmapping` _replaces_ whatever is in the image when the container is launched. But once the container is up, any changes it makes are persisted. To get things working, I launched a django container and re-ran the config generation command. This put `config_settings.py` back in its proper place, and everything has worked normally since then. If other people are able to replicate by manually deleting `config_settings.py` from `/vagrant`, we may need to revisit the config generation so that it's more resilient.

pcaisse:
@ddohler I was able to reproduce the issue you described.

I don't think it's caused by the changes in this PR in particular, so I'm going to merge, but we might want to address this issue.

kshepard:
+1 for a future story to fix that issue. I just ran into it as well, and it took me a while to resolve it.

PR #513, opened 2018-02-15
Fix all tests

ddohler:
## Overview

Fixes all remaining tests!

### Checklist

### Notes

* This doesn't imply that Travis should succeed; it may not (but it shouldn't be the tests' fault).
* However, we should start running tests locally as a matter of course now that they work. This PR was a bit of a moving target because tests were failing out from underneath me each time I updated `develop`.
* There was one skipped test that became unskipped because we made changes to the settings. However, our changes didn't go far enough to get things working, and I wasn't able to quickly fix it, so I changed the skip condition to more specifically target the configuration that the test actually needs, and it's still skipped.

## Testing Instructions

 * `./scripts/server`
 * `./manage.py test`
 * Should succeed


django/publicmapping/redistricting/management/commands/setup.py
```diff
@@ -47,16 +40,9 @@
...
-    Geolevel,
-    Geounit,
-    Subject,
-    Characteristic,
-    LegislativeBody,
-    Plan,
-    ProcessingState,
-    configure_views
-)
+from redistricting.models import (Geolevel, Geounit, Subject, Characteristic,
```
pcaisse, line 43: I'm personally a fan of how it was before (each item in a list on one line). More lines are taken up but it's trivial to make your editor sort them which makes it easier to read and find what you're looking for.

ddohler: I agree -- this is a `yapf` artifact. If this is the price we pay for having automated formatting then I guess I'm okay with it, but if we can configure `yapf` to do multi-line imports I'd be happier. Not sure whether that's possible -- it looks like it's configurable to some extent but it may be fiddly.

pcaisse (approved):
LGTM :white_check_mark:

PR #514, opened 2018-02-19
Fix Leaderboard

pcaisse:
## Overview

Fix template for "My Ranked Plans", plus miscellaneous cleanup.

Also demonstrate that validation is working and is respecting `config.xml`.

### Checklist

### Notes

When we were testing validation before, we were editing the wrong nodes in `config.xml`. It appears that only Geoserver cares about `<ScoreFunction id="a_congressional_population">`. The validation we care about for the `Equipopulation` calculator is `<ScoreFunction id="a_congress_plan_equipopulation_validation">`.

None of this is particularly clear. Seems like we might want to revisit configuration documentation for our own benefit/sanity as well as that of our future users.

### Demo

**Before**
my_ranked_plans_broken

**After**
my_ranked_plans_fixed

## Testing Instructions

 * `./scripts/server`
 * Log in (may need to register first)
 * Assuming you have the full data load, select "Congressional" plan, name the plan, and start drawing
 * Click "Share" tab
 * Click "Verify and Submit Plan to Leaderboards" button
 * Validation should fail with a message. Close the dialog.
  * In `config.xml`, go to the `ScoreFunction` node with an id of `a_congress_plan_equipopulation_validation`. Change the `min` argument to 0 and `max` to some huge number
 * Run `docker-compose exec -T django ./manage.py setup config/config.xml -f` (`force` flag is important so that the arguments are updated)
* Click "Verify and Submit Plan to Leaderboards" button again
* After the data loads, you should see your plan under "Top Ranked Users' Plans" in the leaderboard
* Click the "My Ranked Plans" tab and you should see the same plan listed

Closes #154978264

### DIFF
--- a/django/publicmapping/redistricting/calculators.py
+++ b/django/publicmapping/redistricting/calculators.py
@@ -615,7 +615,7 @@ class PolsbyPopper(CalculatorBase):
     """
     Calculator for the Polsby-Popper measure of compactness.
 
-    The Polsby-Popper measure of campactness measures the area of a circle
+    The Polsby-Popper measure of compactness measures the area of a circle
     with the same perimeter as a district compared to area of the district.
 
     This calculator will calculate either the compactness score of a single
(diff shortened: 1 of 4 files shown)


shreshthkhilani (approved):
This looks good. The only note is that existing plans are still held to the old min and max rules, creating a new plan allows sharing to leaderboard.

PR #517, opened 2018-02-22
Fix SLD overlays not showing up

ddohler:
## Overview

Fixes the problems with colored overlays not showing up that we identified in planning yesterday.

### Checklist

### Demo

screen shot 2018-02-22 at 12 20 29 pm

### Notes

- This was broken in two ways. The first one was that the Django container didn't have a mapping to the SLD directory, so the SLD files it generated weren't making it into the folders that the Nginx and Geoserver containers were using.
- The second problem was fallout from the calculator module path change which prevented some styles from being generated.
- I also discovered and fixed a problem with our get-or-create logic for styles, namely that it wasn't being passed a valid URL. The only thing that I noticed that this broke was that it would try to re-create styles that already existed and receive an error from Geoserver, which it would then warn about but continue.
- Because I converted the SLD folder into a Docker volume, I deleted the `sld` folder that we had previously.

## Testing Instructions

 * `scripts/server`
 * `./manage.py setup config/config.xml -G`
 * You may get an error that the Geoserver configuration could not be cleaned. If this happens, let me know and I'll create a story to resolve it (I know a little bit about why it happens, but haven't solved it completely so I'm not completely sure whether it's a local or global problem). You can proceed with testing by using the GeoServer web interface to delete the `pmp` workspace.
 * View a plan and revel in the gorgeous colored overlays.


pcaisse:
It's beautiful...

sld

pcaisse (approved):

PR #518, opened 2018-02-22
Fix CSRF error when uploading plan files

ddohler:
## Overview

Fixes a CSRF error when uploading plans. Also fixes a second error that was hidden by the first, related to changes to how Celery serializes objects by default.

### Checklist

- ~[ ] Files changed in the PR have been `yapf`-ed for style violations~ (I tried this but it just made two small changes that weren't actually problems, so I skipped it).

## Testing Instructions

 * Log in and go to "Upload Plan"
 * Choose a file and upload it. If you want to see a complete success on the next step, either grab one of the index files from the sample Virginia data zip file, or request an index file for an existing template plan. Otherwise, you can just upload any old file.
 * If you uploaded a plan index file, you should see a note about your file being processed. If you uploaded some other kind of file, you should see a message that the file wasn't able to be processed. You shouldn't see a CSRF error, nor should you see a Serialization Error.


django/publicmapping/redistricting/management/commands/setup.py
```diff
@@ -723,7 +722,7 @@ def create_template(self, config):
                 plan_name,
                 legislative_body.id,
                 path,
-                owner=admin,
+                owner_id=admin.pk,
```
jisantuc, line 725: This breaks the processing in `index2plan`, which is still expecting an `owner`, rather than an `owner_id` -- do you think it's appropriate to carry this change through to celery in this PR, or to make another story and fix it there?

ddohler: I think I've made the necessary changes to `index2plan` in this PR; is there something I've missed? Regardless, that was definitely my intent, so if there's anything else that needs to be done I can add it.

jisantuc: Hmm, lemme rebuild containers and see if that magic fixes it...

jisantuc: Ok yeah thanks rebuilt containers -- now getting an email about not being able to send emails, which, yeah, DistrictBuilder, I _know_

jisantuc (approved):

PR #519, opened 2018-02-22
Fix sharing of plans

pcaisse:
## Overview

Fix plan sharing.

### Checklist

## Testing Instructions

 * `./scripts/server`
 * Log in as registered user (not guest)
 * Choose a template plan assuming you have some loaded
 * Give it a name and start drawing
 * Click the "Share" tab
 * Give it a name and click "Save and Share"
 * Note that it works (success dialog is shown, newly generated link/plan works)
 * Click "Plan" tab
 * Click "My Plans" and select a plan
 * Choose "Copy and save selected plan", give it a name, and start drawing
 * Go back to "Plan" tab and notice that that plan is _not_ shared

### DIFF
--- a/django/publicmapping/redistricting/views.py
+++ b/django/publicmapping/redistricting/views.py
@@ -298,7 +298,8 @@ def copyplan(request, planid):
     newname = p.name + " " + str(random.random())
     if (request.method == "POST"):
         newname = request.POST["name"][0:200]
-        shared = request.POST.get("shared", False)
+        shared = request.POST.get("shared",
+                                  False) and request.POST["shared"] == "true"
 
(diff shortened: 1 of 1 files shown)


ddohler (changes requested):
I'm having trouble getting this to work; when I try to "Start Drawing", Django throws a 500 error with the message `IntegrityError: null value in column "is_shared" violates not-null constraint` . That looks like it's the same error message that's showing up in the tests, so I bet the fix is the same for both.

pcaisse:
I did some hasty refactoring EOD yesterday... should be good now!

ddohler (approved):
Works great!

PR #521, opened 2018-02-26
Make Geoserver admin user password configurable #155082923

pcaisse:
## Overview

The Geoserver admin username/password is by default admin/geoserver. This is easily discoverable online, so we want to change it as part of the setup process to what's defined in `.env`.

### Checklist

### Notes

The script assumes that the password is 'geoserver' for it to be reset. As such, the idea is for this password to only need to be reset once (subsequent resets won't do anything).

As with other components in the system, if you change passwords in `.env` without starting over, things may not work right or steps may have to be repeated. This "foward only" assumption seems to have been in place up until now, so I don't think it hurts to keep going in that same vein.

## Testing Instructions

 * In `.env`, change `MAP_SERVER_ADMIN_PASSWORD` to something besides 'geoserver'. Delete `MAP_SERVER_ADMIN_USER` from `.env`.
 * Run `./scripts/update` (password will be reset based on `.env` by running this script)
 * `./scripts/server`
 * Go to http://localhost:8080/geoserver/web/ and log in using the new password.
 * `vagrant ssh` and `docker-compose exec -T django ./manage.py setup config/config.xml -G`
 * Configuration of Geoserver should succeed as before

Closes #155082923


geoserver/change_admin_password.sh
```diff
@@ -0,0 +1,34 @@
...
+# This nees to be reset to provide some minimum level of security.
+# -----------------------------------------------------------------------------
+
+echo "Attempting to set a new admin password..."
+echo "Note that this assumes the current password is the default ('geoserver')"
+
+# Sometimes geoserver is not ready to receive API calls for whatever reason.
+# Retrying seems to do the trick.
+n=0
+until [ $n -ge 5 ]
```
ddohler: This is fine, but I think it could be shortened by using `for i in {1..5}`.

geoserver/change_admin_password.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# -----------------------------------------------------------------------------
+# Attempt to reset the admin user's password.
+#
+# By default, the admin username is 'admin` and the password is 'geoserver'.
+# This nees to be reset to provide some minimum level of security.
+# -----------------------------------------------------------------------------
+
+echo "Attempting to set a new admin password..."
```
ddohler: When running this from within `./scripts/update` I noticed that there's no context about what password it's trying to set, except for the message about the default on the next line. It'd be good to clarify that this is setting the geoserver admin password.

ddohler (approved):
This is good for now, but I think long-term we may need to explore other approaches because the net result of this is that we have to re-run the password script every time we launch a new container from the image (i.e.  `docker rm vagrant_geoserver_1` will reset to the default password next time you do `./scripts/server`). I think that's fine as long we're in a one-machine deployment scenario, but if we ever move this to something like ECS or EKS it could become more problematic. I noticed that Geoserver can use a [JDBC authentication provider](http://docs.geoserver.org/latest/en/user/security/webadmin/auth.html#jdbc-provider), which would essentially cause Geoserver to delegate its authentication process to a database server. That might be one possibility for making this more easily manageable. @tnation14 do you have any other thoughts?

tnation14:
It looks like GeoServer gives you the ability to configure users via a `users.xml` file in `$GEOSERVER_DATA_DIR` ([source](http://docs.geoserver.org/stable/en/user/security/usergrouprole/usergroupservices.html#xml-user-group-service)). I think it would be good to have `users.xml.template`, and then use `envsubst` to create `$GEOSERVER_DATA_DIR/security/usergroup/<name>/users.xml` (simliar to the process for creating `nginx` conf files). That way, the admin user is automatically configured when the server starts, and doesn't need to be programmatically reset at all.

ddohler:
I looked into that, but it seemed like it would be tough to get the password into the format that geoserver is looking for; the `crypt1` encryption [runs MD5 on the password](http://docs.geoserver.org/stable/en/user/security/passwd.html#password-based-encryption) 1000 times to generate a key and then uses that key to encrypt the password using DES. I bet we could probably replicate the process using some of the OpenSSL command-line tools, but we'd still have some kind of pre-processing step required prior to using `envsubst`. My thought was that JDBC would be easier since Geoserver already knows where to find Postgres, but it might also be possible to add a helper script that would run the encryption that people can use when setting up their `.env` files.

pcaisse:
> the admin user is automatically configured when the server starts, and doesn't need to be programmatically reset at all

@tnation14 That was the road I went down initially, but I ran into problems. I don't see a way to easily replace the encrypted password in the XML file without reimplementing the encryption method (of which there are several). I also am not clear on potential syncing issues (if we change the XML, how do we make sure that that file and the database are in sync). There doesn't seem to be a lot of info on that in the docs and they don't seem to encourage you to that in any way. Going against the grain there seemed like it would be asking for trouble. The docs actually seem to push you to configure users manually through the UI, but that defeats the purpose of automating password reset. Using the API to interact with the service as a black box where we just make a request and if it's successful it takes care of the encryption, storage, and whatever else seemed like the best option.

tnation14:
>  My thought was that JDBC would be easier since Geoserver already knows where to find Postgres

@ddohler This is a really good point. JDBC was my second choice, but it's probably easier to implement than reproducing the encryption method. Out of the two remaining authentication methods, I'm leaning more towards this one.

>Using the API to interact with the service as a black box where we just make a request and if it's successful it takes care of the encryption, storage, and whatever else seemed like the best option.

@pcaisse yeah, I think what's in this PR definitely beats trying to replicate the encryption setup; my thought process here is just that I'd prefer to have services configured at runtime rather than being manually set after-the-fact.

tnation14 (approved):

PR #523, opened 2018-02-27
Use correct attributes for Geoserver feature creation

ddohler:
## Overview

Fixes feature creation for Geoserver. Also fixes Reference Layers (perhaps). The main problem was that we were passing the wrong set of attributes to Geoserver when trying to create features based on our table views. The attributes for each feature must be a subset of the columns in the table or view. I resolved most of the issues by using the database console to inspect the views corresponding to each failed feature, and then changing that feature's attribute list to match.

### Checklist

- [X] Files changed in the PR have been `yapf`-ed for style violations

### Notes

- I'm not sure if the Reference Layer functionality is completely fixed; it doesn't result in angry red broken tiles anymore, but not all of the options seem to have a noticeable effect on the map. I didn't dig much further but this may need to be addressed if it's not expected functionality.
- This also significantly simplifies the code for clearing out the existing Geoserver configuration by using the `recurse` flag.

## Testing Instructions

From inside the VM
 * `./scripts/server`
 * `./manage.py setup config/config.xml -G`
 * It should work without any error messages

Closes #155399854

### DIFF
--- a/django/publicmapping/redistricting/config.py
+++ b/django/publicmapping/redistricting/config.py
@@ -1252,26 +1252,25 @@ class SpatialUtils:
             return False
 
         # Create the feature types and their styles
-
-        geolevel_fields = [
-            x for x in Geolevel._meta.fields
-            if x.get_internal_type() != 'ForeignKey'
-        ]
-        geolevel_attrs = [{
-            'name': field.name,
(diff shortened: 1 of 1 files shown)


pcaisse (approved):
LGTM

PR #526, opened 2018-03-01
Make info-click functionality transfer less data [#155400639]

ddohler:
## Overview

This improves the functioning of the info-click feature. Previously, the function was making a request that resulted in a multi-megabyte Geoserver response containing hundreds of geometries. This was slowing down both Geoserver and frontend handling of the data, and may have also resulted in errors (I saw one response that appeared to be truncated, although that could have been the developer tools' fault). Now, it makes one request for only the necessary attributes to do front-end filtering; once the filtering has been completed, a second request is made for a single geometry.

This also includes some refactoring to simplify the code.

### Checklist

- ~[ ] Files changed in the PR have been `yapf`-ed for style violations~

### Notes

Although I'm fairly certain that the info-click feature is now reliably working _as designed_, I wouldn't say it works _well_. Multiple clicks within the same _visible_ boundary can return different sets of demographics, or no demographics at all. This is because the set of Geounits returned by the Geoserver query changes depending on where the map is clicked. It's not currently clear at what level the issue lies; it could be caused by interactions between any of these stages:
- Source data or configuration
- Data loading
- The logic which generates the `identify_geounit` view
- The frontend filtering logic

I've written up a story for someone to take a more holistic look at the process and see if they can figure out what's going on.

Also, the two-request approach is not always the best one; for smaller geometries the overhead of making two requests exceeds the slowness caused by handling many geometries. However, in most cases, the two-request approach is faster. After having looked into this a bit, I think it should be possible to avoid doing any client-side filtering at all, and instead make a Geoserver request that returns only the exact Geounits that are needed. This would be the ideal solution if we need to make this faster. Additionally, the geometry data is only used to calculate the centroid of the geometry in order to determine where to put the pop-up. We could make this even faster by pre-calculating the centroid on the backend and using that in the `identify_geounit` view rather than including the whole geometry.

## Testing Instructions

 * Before checking out this PR, click around the map on `develop` with the info button to get a sense of how long it takes and where the "dead zones" are. If you get the "No demographics" modal, keep clicking near the same area to see if you can get a modal in some other part of the boundary. Open the dev tools and see that the WFS requests have large response sizes.
 * Then check out this branch and repeat. The behavior should be the same, or possibly with fewer dead zones, and it should also be somewhat faster in most cases. In the dev tools there should be two requests for each click, rather than one, but they should both be much, much smaller than on `develop`.


pcaisse:
I see this working on `develop` but not on your branch. On your branch, I get this as a response from the `wfs` endpoint:
```<?xml version="1.0" ?>
<ServiceExceptionReport
   version="1.2.0"
   xmlns="http://www.opengis.net/ogc"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wfs/1.0.0/OGC-exception.xsd">
   <ServiceException code="InvalidParameterValue">
      Requested property: id is not available for pmp:identify_geounit.  The possible propertyName values are: [name, geolevel_id, geom, number, percentage, subject_id]
</ServiceException></ServiceExceptionReport>```

ddohler:
Oops, yeah, that was my fault -- I had the `id` attribute in there as part of my work on the Geoserver errors, but I took it out at the last second and didn't rerun configuration afterward, so my Geoserver had it but no one else's did. I've pushed up a commit to fix that; you'll just need to run `./manage.py setup config/config.xml -G`.

pcaisse (approved):
Working now and is much faster :+1:

PR #528, opened 2018-03-02
"Fix" geolevel descriptions

jisantuc:
## Overview

This PR improves the default translation string for regional geolevel long descriptions
and does minimal prep work to get the admin site working. It also excludes the `locale`
files from `vagrant rsync` in preparation for what's going to come next to get translations
working appropriately.

### Checklist

### Notes

It turned out that there was nothing wrong with the labels, just our understanding of how they
get there. When we import geolevels, we try to populate message strings for each language
in the settings file. If we can't make a reasonable guess, we just throw in an empty string.

Then, when we try to get translated strings, if we find an empty string, we fall through to the
message id, which is something like `va_county vtd long description`. The actual fix is going
to require a strategy to make changes to the language files in the VM.

@kshepard says that we can use rosetta to do this. We have
[rosetta installed](https://github.com/PublicMapping/DistrictBuilder/blob/feature/js/fix-geolevel-descriptions/django/publicmapping/requirements.txt#L9),
but it depends on the admin site (which we can't currently get to) and possibly on some
additional configuration.

I've opened some PT stories to track admin work

## Testing Instructions

 * Bring up your server
 * Reload the lowest resolution geolevel -- `docker-compose exec django ./manage.py setup config/config.xml -g2`
 * Go to your map
 * Confirm that what used to say `va_county long description` now just says `county` (see the
   [PT story](https://www.pivotaltracker.com/story/show/155399721) for screenshot)
 * Edit `django/publicmapping/locale/en/LC_MESSAGES/xmlconfig.po` outside the vm:
   * find the `msgid` for `va_county long description`
   * change the `msgstr` to something else and save
   * reload the django app and check that the label *has not* changed


ddohler (approved):
Works great! I had to reload the Django app to get the `county` label to show up after loading the g2 level, but that was the only hiccup.

PR #532, opened 2018-03-05
Fix study area extent calculation

kshepard:
## Overview

The study area extent calculation was very inefficient, particular if your base geolevel has a large number of geounits. In a case, such as PA municipalities, the page would time out while attempting to load. This PR uses an `Extent` aggregation to speed things up.

### Checklist

## Testing Instructions

 * Load the map page (this will probably be a little faster than normal)
 * That's it. If this code fails, an error will be immediately observable and/or nothing will work, since the extent is used to determine many fundamental things (such as map zoom levels).


django/publicmapping/redistricting/models.py
```diff
@@ -557,6 +557,15 @@ def renest(self, parent, subject=None, spatial=True):
...
         return True
 
+    def calc_extent(self):
+        """
+        Returns the extent (array of four floats) of the geounits belonging to this geolevel
+        """
+        extent = self.geounit_set.aggregate(Extent('simple'))['simple__extent']
+
+        # Convert the 4-tuple returned by the query to a list
+        return [extent[0], extent[1], extent[2], extent[3]]
```
pcaisse: Could also do `return list(extent)`

pcaisse (approved):

django/publicmapping/redistricting/models.py
kshepard: Wow, I totally didn't realize that. Thanks, that's much better.

PR #537, opened 2018-03-14
Provide documentation on how to do translations #155663213

pcaisse:
## Overview

Provide documentation on how to do translations

## Testing Instructions

 * Follow instructions in documentation to make a translation via Rosetta and get it onto the host

Closes #155663213

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -129,7 +129,7 @@ Once you are happy with your changes, the next step is to get them out of the VM
 You can use the command `vagrant ssh-config` to find the host, port, user, and identity file -- all of which you will need to copy the files over -- and then run:
 
 ```
-scp -i <IdentityFile> -P <Port> -r <User>@<Host>:/vagrant/django/publicmapping/locale/ django/publicmapping/locale/
+scp -i <IdentityFile> -P <Port> -r <User>@<Host>:/vagrant/django/publicmapping/locale/ django/publicmapping/
 ```
 
 You can then verify the translations are correct and commit those files.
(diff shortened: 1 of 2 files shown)


README.markdown
```diff
@@ -117,6 +117,23 @@ that the script executes:
...
+DistrictBuilder uses a Django application called [Rosetta](https://github.com/mbi/django-rosetta) to do translations.
+
+To do translations in a given language, make sure the server is running (`./scripts/server`) and go to `/admin` to log in as an admin (the admin username and password are defined in `.env`). Once logged in, go to `/rosetta`. You should see the different languages available and the paths of the files that correspond to each language. If you make a translation and save, you should see your changes in that file in the `django` container and on the VM. If you restart the server, your translations will be visible in DistrictBuilder.
+
+Once you are happy with your changes, the next step is to get them out of the VM and onto the host so they can be saved in source control.
+
+You can use the command `vagrant ssh-config` to find the host, port, user, and identity file -- all of which you will need to copy the files over -- and then run:
+
+```
+scp -i <IdentityFile> -P <Port> -r <User>@<Host>:/vagrant/django/publicmapping/locale/ django/publicmapping/locale/
```
kshepard: The last piece of this command should be `django/publicmapping/` rather than `django/publicmapping/locale/`, as the first version will create an additional `locale` subdirectory.

kshepard (approved):
This looks good after fixing up the one thing I mentioned. Another thing I discovered while testing is that `Rosetta` is not picking up on the `xmlconfig.po` files. That shouldn't prevent this PR from being merged, but is worth looking into as part of the broader `Rosetta`-related task.

README.markdown
pcaisse: Good catch!

PR #538, opened 2018-03-21
Fix info click demographic info popup

pcaisse:
## Overview

Fix info click demographic info popup.

### Checklist

### Notes

This issue was initially identified when working on making demographic info load faster (see "Notes" in https://github.com/PublicMapping/DistrictBuilder/pull/526).

When beginning to work on this bug it was unclear where the problem lay as we had made changes to how data was loaded somewhat recently. It could also have been due to misconfiguration of our VA data, an issue with our database, or with our filtering on the frontend.

After investigating, it seems to be the frontend filtering. Previously, we were requesting data from Geoserver only with a spatial filter for where the user clicked without filtering by geolevel. Then we would get features back from a variety of geolevels. Depending on how far the user was zoomed in or out, the acceptable geolevels would be highly restricted (in the case of VA data, all geolevels except 6 are filtered out when fully zoomed out). Oftentimes a feature with a geolevel of 6 would not be returned in the exactly 101 features returned, in which case no information was shown in the popup, or only some information was returned.

The solution to this issue is to filter by geolevel on the backend so that we only get features for the geolevels we care about. 

## Testing Instructions

 * On `develop`, with the app running go to a plan.
 * Click the "Info" button under "Map Editing" on the "Draw" page.
 * Zoom in somewhat and click around in a county and notice that you get different information in the pop up as you click around (either all, some, or none). Zoom out fully and notice that after clicking you often get no info. 
 * Checkout out branch and refresh page. You should always get all the info at any zoom level wherever you click. Check that the more you zoom in, the more finer grained the information is (eg. zooming in fully in Stanton, VA will get you info at the block level).

Closes #155625445

### DIFF
--- a/django/publicmapping/static/js/mapping.js
+++ b/django/publicmapping/static/js/mapping.js
@@ -152,6 +152,9 @@ function init() {
 
 // Timer used to distinguish between single and double clicks
 var clickTimer = null;
+// Min and max geolevels used for filtering requests to Geoserver
+// and for display of demographic info popup.
+var geolevelRange;
 
 /**
  * Toggles the highlighting of a district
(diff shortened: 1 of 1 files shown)


ddohler (approved):
I have the PA data loaded, which I don't think has enough geolevels to exhibit the problem, so I wasn't able to confirm that this fixes it, but I did confirm that this works as expected.

PR #546, opened 2018-05-01
Replace Ansible provisioning with a `docker-compose` based deployment

### LINKED ISSUE #542: Deploy DistrictBuilder service using docker-compose and `DOCKER_HOST`
#### Problem description

Our current deployment pipeline deploys images to quay.io, and uses a combination of Ansible and Terraform to provision the AppServer with an Upstart service that runs `docker-compose  up`. This process is complex, and can be difficult for new engineers to a) learn and b) debug when it breaks. Additionally, since this Git repo is the only one that deploys to our Quay repository, users cannot deploy their own DistrictBuilder forks (i.e. [azavea/district-builder-dtl-pa](https://github.com/azavea/district-builder-dtl-pa)) unless we build/publish a container image for their fork. Finally, the logging and automatic restart behavior managed with Upstart service can instead be managed by `docker-compose`.

#### Replication Steps

```
$ docker-compose -f docker-compose.ci.yml run --rm terraform ./scripts/infra plan
$ docker-compose -f docker-compose.ci.yml run --rm terraform ./scripts/infra apply
```

#### Expected Behavior or Output

Instead of using Ansible to create an Upstart service tied to a specific docker image, we should:
- Create `docker-compose.production.yml` based on the one at [deployment/ansible/roles/district-builder.app-server/templates/docker-compose.yml.j2](https://github.com/PublicMapping/DistrictBuilder/blob/develop/deployment/ansible/roles/district-builder.app-server/templates/docker-compose.yml.j2)
    - Use Compose's `restart_policy` configuration key to control restart behavior
    - Use Compose's `logging` configuration key to send all logs to `/var/log/districtbuilder.log`
    - Remove references to `quay.io/azavea/districtbuilder-*` images
- Use the `DOCKER_HOST` environment variable to build images directly on the app server and start the service (see [Docker docs](https://docs.docker.com/engine/reference/commandline/cli/#environment-variables))

This streamlines the deployment process to running `docker-compose build` and `docker-compose up` after Terraform runs, and removes the need to host images on Quay.

### LINKED ISSUE #544: Update external deployment documentation
#### Problem description

The terraform template 'sample.tfvars' referenced in deployment/README.md does not exist. It would be helpful to have this when creating new deployments.

tnation14:
## Overview

This PR replaces complicated Ansible tooling with `docker-compose` based deployments to simplify the deployment process and decouple application deployment from Azavea's AWS infrastructure. Now, users will be able to deploy the application on the infrastructure provider of their choice without being tied to our internal deployment scripts. Additionally, I added a `Deployment` section to the project `README` to separate the general deployment instructions from internal ones.

- Remove ansible provisioning files
    - The App server instance now installs dependencies and does pre-deploy setup using `cloud-init`.
- Add `docker-compose.production.yml`
    - `docker-compose.yml` defines the base architecture for the application
    - `docker-compose.overrides.yml` contains development environment-specific configurations (volume mounts, `postgres` container, etc).
    - `docker-compose.production.yml` contains Prod environment-specific configurations  (send all logs to syslog, `on-failure` restart policy)
- Remove unnecessary STRTA (`scripts/cibuid`, `scripts/cipublish`)
- Add remote `docker-compose` based deploy
    - Install `docker-compose` in Terraform container
    - Add additional AppServer security groups to allow Docker access from CI
    - Secure Docker by enabling TLS
    - Add a `--production` flag to development STRTA for cross-environment functionality
    - Modify `scripts/infra` to check if docker is running, then build images and start services directly on the AppServer instance
- Update deployment documentation

### Checklist

- [ x ] PR has a name that won't get you publicly shamed for vagueness
- [ ] ~Files changed in the PR have been `yapf`-ed for style violations~

### Demo

https://va.districtbuilder.azavea.com

### Notes
- For more info on Docker over HTTPS, see [Docker Docs](https://docs.docker.com/engine/security/https/)
- Docker server TLS certificates are stored at `s3://districtbuilder-staging-config-us-east-1/docker_certs/server/{pa,va}.zip`. Certificates for other instances will need to be created manually and stored in the same location.
- Docker client TLS certificates are stored at `s3://districtbuilder-staging-config-us-east-1/docker_certs/client/client.zip`. These are general purpose and do not need to be created on a per-state basis.

## Testing Instructions

 * I modified `.travis.yml` to deploy this branch to staging. See [Travis Console](https://travis-ci.org/PublicMapping/DistrictBuilder/builds/373527103?utm_source=github_status&utm_medium=notification)
* See https://va.districtbuilder.azavea.com

Closes #542 
Closes #544


README.markdown
```diff
@@ -1,23 +1,37 @@
...
+    * [Requirements](#deployment-requirements)
+    * [Setup](#setup)
+    * [Setting Up the Production Environment](#setting-up-the-production-environment)
 
-Installation
-------------
 
-### Requirements ###
+# Installation
+<h3 id=installation-requirements>
```
hectcastro: These IDs happen automatically with Markdown headers. The ID value maps to the name of the heading. In this example, it would be `#installation`.

tnation14:
Needs https://github.com/azavea/district-builder-core-deployment/pull/2. Also: I added some documentation about creating TLS certificates [here](https://github.com/azavea/district-builder-core-deployment/wiki/Generating-Docker-TLS-certificates).

README.markdown
```diff
@@ -152,3 +165,96 @@ $ pre-commit install
...
+
+<h3 id=deployment-requirements>
+Requirements
+</h3>
+
+- Docker v17+
+- Docker-Compose 1.21+
+- Git
+- PostgreSQL server 9.5
+  - PostGIS 2.2
```
hectcastro: Extra space.

README.markdown
```diff
@@ -1,23 +1,37 @@
...
+    * [Development Environment Setup](#development-environment-setup)
+      * [`config.xml`](#configxml)
+      * [Setting Up Your Application](#setting-up-your-application)
+  * [Translations](#translations)
+  * [Support](#support)
+  * [Development](#development)
+  * [Deployment](#deployment)
+    * [Requirements](#deployment-requirements)
+    * [Setup](#setup)
+    * [Setting Up the Production Environment](#setting-up-the-production-environment)
```
hectcastro: Nitpick, but this section and the one above it both refer to "Setup". I think that the one after "Setup" could be labeled "Configuration" or something similar.

deployment/terraform/templates/cloud-config.tpl
```diff
@@ -0,0 +1,18 @@
...
+  - touch /opt/district-builder/user-data/config_settings.py
+  - chown -R ec2-user:ec2-user /opt/district-builder
+  - touch /home/ec2-user/permissions-ready
+  - usermod -aG docker ec2-user
+  - yum update -y
+  - yum groupinstall -y 'Development Tools'
+  - yum install -y libffi-devel
+  - curl 'https://bootstrap.pypa.io/get-pip.py' -o '/tmp/get-pip.py'
+  - python /tmp/get-pip.py
+  - /usr/local/bin/pip install docker-compose==1.16.* awscli
```
hectcastro: I believe the AWS CLI is already installed on ECS AMIs.

hectcastro: Also, you can install the Docker Compose binary directly via `curl` without needing to install `pip` first:

```
$ sudo curl -L https://github.com/docker/compose/releases/download/1.21.0/docker-compose-$(uname -s)-$(uname -m) -o /usr/local/bin/docker-compose
```

See: https://docs.docker.com/compose/install/#install-compose

deployment/terraform/templates/cloud-config.tpl
```diff
@@ -0,0 +1,18 @@
...
+  - /usr/local/bin/aws s3 cp s3://${zip_file_uri} /tmp/certs.zip
+  - unzip -d /etc/docker/certs /tmp/certs.zip
+  - echo 'OPTIONS="$${OPTIONS} -H unix:///var/run/docker.sock  -H 0.0.0.0:2476 --default-ulimit nofile=1024:4096 --tlsverify --tlscacert=/etc/docker/certs/ca.pem --tlscert=/etc/docker/certs/server-cert.pem --tlskey=/etc/docker/certs/server-key.pem"' >> /etc/sysconfig/docker
```
hectcastro: This might be easier to manage with `write_files`.

deployment/terraform/compute.tf
```diff
@@ -54,96 +42,25 @@ resource "aws_instance" "app_server" {
...
-      bastion_private_key = "${file(pathexpand("~/.ssh/district-builder.pem"))}"
-    }
-  }
-
-  provisioner "file" {
-    source      = "${path.root}/../ansible/district-builder-app-server.yml"
-    destination = "/opt/district-builder/deployment/ansible/district-builder-app-server.yml"
+  provisioner "remote-exec" {
+    inline = [
+      "while [ ! -f /home/ec2-user/permissions-ready ]; do sleep 2; done",
```
hectcastro: Could this wait be eliminated by just synchronizing to a known location, then copying from there to `/opt/district-builder/user-data/` after it is created?

scripts/cipublish
```diff
@@ -1,37 +0,0 @@
...
-if [ "${BASH_SOURCE[0]}" = "${0}" ]
-then
-    if [ "${1:-}" = "--help" ]
-    then
-        usage
-    else
-
-    docker login -u "${QUAY_USER}" -p "${QUAY_PASSWORD}" quay.io
-
-    docker push "quay.io/azavea/districtbuilder-nginx:${GIT_COMMIT}"
```
hectcastro, line 31: Reminder to delete these images from Quay once they're no longer being used, or create an issue so that we can remember to do it later.

scripts/infra
```diff
@@ -41,12 +58,34 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
...
+                    aws s3 cp "s3://${DB_SETTINGS_BUCKET}/docker_certs/client/client.zip" "client.zip"
+                    unzip -od "/root/.docker" "client.zip"
+                    rm "client.zip"
+
+                    export DOCKER_TLS_VERIFY=1
+                    export DOCKER_HOST="origin.${DB_STATE}.districtbuilder.azavea.com:2476"
+                    export DOCKER_CERT_PATH="/root/.docker"
+
+                    set +e
+                    wait_for_docker
```
hectcastro, line 83: Is this mostly to cover the edge case where a new machine is coming online and we're account for the Docker daemon restarting?

.travis.yml
```diff
@@ -16,14 +16,14 @@ install:
 - cp django/publicmapping/config/config.dist.xml django/publicmapping/config/config.xml
 - "./scripts/update"
 script:
-- "./scripts/cibuild"
+- "echo 'ran a script'" # "./scripts/test"
```
hectcastro: Reminder to update this later.

scripts/load_configured_data
```diff
@@ -52,7 +59,15 @@ then
         usage
     else
         configPath=${1:-config/config.xml}
-        docker-compose up -d postgres django
+        if [ "${1:-}" = "--production" ]
+        then
+          export COMPOSE_FILE="docker-compose.yml:docker-compose.production.yml"
```
hectcastro: It is unclear why variables are being exported here. Is the idea that they need to be available to `docker-compose` for one condition and not the other?

scripts/infra
tnation14: That's correct.

deployment/terraform/templates/cloud-config.tpl
tnation14: The reason I opted for this approach over `write_files` was that this allows us to specify our own Docker options without overwriting the `/etc/sysconfig/docker` file that comes with the ECS AMI. I didn't want to get into a situation where we're overwriting Amazon-specific docker settings.

deployment/terraform/compute.tf
tnation14: I don't think so. The file provisioner runs as soon as SSH is available, so it runs at the same time as cloud-init. Without a wait/signal, we run the risk of trying to move the file into the user-data directory before a) it exists or b) the permissions have been set.

scripts/load_configured_data
tnation14: That's correct, yeah. I'll add a comment above that line to clarify.

deployment/terraform/templates/cloud-config.tpl
tnation14: > I believe the AWS CLI is already installed on ECS AMIs.

I just launched an instance based on `ami-aff65ad2` (latest ECS AMI), and I actually don't think it is:
```bash
[ec2-user@ip-10-0-2-133 ~]$ which aws
/usr/bin/which: no aws in (/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/home/ec2-user/.local/bin:/home/ec2-user/bin)
```
Since both packages have to be installed anyway, I'm inclined to leave the pip install as is.

scripts/cipublish
tnation14: Created #547

README.markdown
tnation14: Fixed. I initially used the HTML syntax because `#requirements` was clashing between the Development and Production `Requirements` sections. Turns out that when there's duplicate headings, GitHub Markdown appends a `-1`, `-2`, etc to the heading names. So the Deployment `#requirements` ID becomes `#requirements-1`.

README.markdown
tnation14: Fixed. I changed the title of "Setup" to be more accurate, and fixed "Setting Up the Production Environment" to draw a parallel between this section and the development setup section.

deployment/terraform/templates/cloud-config.tpl
hectcastro: I tested with a RF container instance and that had it installed. Its existence on those instances may have been influenced by the installation of `awslogs` via `cloud-config` `packages` though.

I still think that it would be worth installing the `aws-cli` package via `packages`, and if that works, download the Docker Compose binary via `curl`. That approach helps avoid needing `pip` as a dependency, as well as all of the build dependencies necessary.

```yaml
packages:
 - aws-cli
```

In looking through reference material about `packages`, I also ran into `users` and `groups`. I think that we could use them to manipulate `ec2-user` membership of the `docker` group.

```yaml
users:
  - default
  - name: ec2-user
    groups: docker
```

I realize that so much emphasis on funneling working approaches through `cloud-config` may feel valueless, but I think there is long-term value in trying to reuse those directives vs. rolling our own.

deployment/terraform/compute.tf
hectcastro: But if we just synchronized to a directory like `/home/ec2-user` (and just setup everything there), all of the permissions would be set by default. That would allow us to avoid the need to coordinate for file permission setting, right?

In this scenario, I'm desperately trying to find a way to eliminate the potential for deadlock in the event that some failure prevents the only condition in that infinite loop from being met. I think we could get clever and add a safety valve condition too, but seems best to just get rid of it entirely, if possible.

hectcastro (approved):
A large set of changes, but I really think this helps reproducibility (for us and others). Great job. 👍

